PR #1, opened 2017-05-02
Add Pulse listener

leplatrem:
@magopian r?

### DIFF
--- a/pulseworker/main.py
+++ b/pulseworker/main.py
@@ -1,3 +1,4 @@
+import datetime
 import functools
 import logging
 import os
@@ -9,8 +10,9 @@ from mozillapulse.consumers import NormalizedBuildConsumer
 
 
 DEFAULT_SERVER = "https://kinto-ota.dev.mozaws.net/v1"
-DEFAULT_BUCKET = "build-hub-test"
+DEFAULT_BUCKET = "build-hub"
 DEFAULT_COLLECTION = "builds"
+PULSE_TOPIC = "build.#"
 PULSEGUARDIAN_USER = os.getenv("PULSEGUARDIAN_USER", "")
 PULSEGUARDIAN_PASSWORD = os.getenv("PULSEGUARDIAN_PASSWORD", "")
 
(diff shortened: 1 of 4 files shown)


pulseworker/main.py
```diff
@@ -0,0 +1,116 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def epoch2iso(timestamp):
+    dt = datetime.datetime.utcfromtimestamp(timestamp)
+    # XXX: Python 3
+    if sys.version_info >= (3, 2):
+        dt = dt.replace(tzinfo=datetime.timezone.utc)
+    return dt.isoformat()
```
magopian: Nothing quite like some good ol' datetime/timezone voodoo <3

pulseworker/main.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def pulse2kinto(body, msg, client):
+    # https://github.com/mozilla/pulsetranslator#routing-keys
+    buildinfo = body["payload"]
+
+    skip = (
+        buildinfo["tree"] in ("autoland", "try") or
+        buildinfo["buildtype"] in ("pgo",) or
+        buildinfo["tree"] in ("mozilla-inbound",))
```
magopian: are you expecting some more "skip" in the future for the buildtype and the tree? It feels a bit weird to have this `in` with a single-element list

pulseworker/main.py
```diff
@@ -0,0 +1,116 @@
...
+    if skip:
+        logger.debug("Skip routing key '%s'" % body["_meta"]["routing_key"])
+        return  # Do nothing.
+
+    # XXX: no safer way to obtain version?
+    filename = buildinfo["buildurl"].split('/')[-1]
+    version = filename.split('.' + buildinfo["locale"])[0].replace(buildinfo["product"] + '-', '')
```
magopian: Maybe add a comment with an example filename (eg `firefox-55.0a1.en-US.win32.zip`) so it's easier to understand what this piece of code does?

Would it make sense to have a `version_from_filename` helper?

pulseworker/main.py
```diff
@@ -0,0 +1,116 @@
...
+        default_server=DEFAULT_SERVER,
+        default_bucket=DEFAULT_BUCKET,
+        default_collection=DEFAULT_COLLECTION)
+
+    args = parser.parse_args(sys.argv[1:])
+
+    cli_utils.setup_logger(logger, args)
+
+    logger.info("Publish at %s/buckets/%s/collections/%s/records" % (
+        args.server, args.bucket, args.collection))
```
magopian: Or you could use `"Publish at {server}/buckets/{bucket}/collections/{collection}".format(**args)` :)

leplatrem:
> I love it, clear, concise, straight to the point, and awesomely well executed, AMAZING!

❤️

pulseworker/main.py
glasserc: Personally, I find this kind of code very scary. Could we use `arrow` or something instead?

leplatrem: No need for that, once pulse will be compatible with Python 3 we'll remove the `if`

PR #6, opened 2017-05-05
Use fixtures Kinto collection, update & fix decoders.

n1k0:


ui/Decoder.elm
```diff
@@ -13,7 +13,7 @@ buildRecordDecoder =
         |> required "build" buildDecoder
         |> required "download" downloadDecoder
         |> required "source" sourceDecoder
-        |> required "systemaddons" (nullable (list systemAddonDecoder))
+        |> required "systemaddons" (maybe (list systemAddonDecoder))
```
magopian: What's the point of making it `maybe` instead of `nullable`? Isn't the result the same in this context?

magopian:
I guess the only thing faked here is the revision, and I guess this would have been the key to use to "merge" those system-addons information in the current data we have from pulse...

ui/Decoder.elm
n1k0: Using `nullable` here makes decoding fail with:

```
An error occured while fetching the build records: ServerError 200 "OK" "failed decoding json: I ran into the following problems at _.data[87].systemaddons:\n\nExpecting null but instead got: [{\"id\":\"webcompat@mozilla.org\",\"builtin\":\"1.0\"},{\"id\":\"firefox@getpocket.com\",\"builtin\":\"1.0.5\"},
```

I didn't investigate too much why, sorry.

n1k0:
> I guess the only thing faked here is the revision, and I guess this would have been the key to use to "merge" those system-addons information in the current data we have from pulse...

What do you suggest as a fix, should I update something?

ui/Decoder.elm
magopian: ah, it seems it's because the `SystemAddon` decoder isn't correct: It's expecting a `builtin_version` and a `update_version` but there's none of those in the data.

I guess the decoder should instead be:
```elm
systemAddonDecoder : Decoder SystemAddon
systemAddonDecoder =
    decode SystemAddon
        |> required "id" string
        |> optional "builtin" string ""
        |> optional "updated" string ""
```

magopian:
> What do you suggest as a fix, should I update something?

Nope, I think there's no way to improve that, at least until we find a way to "link" pulse data and system-addons data.

ui/Decoder.elm
magopian: I believe the other similar changes were for similar issues? Maybe we also need to change the decoders in those cases?

n1k0: But the patch currently does that already. Switching to `nullable` keeps the error occuring even with the decoder properly updated.

n1k0: Ah I missed the use of optional. Will update.

n1k0: But now I think more about it, don't we have an overall schema issue here? I mean, does it make sense to have a `SystemAddon` entry with no `builtin` and/or `updated` fields? Sounds like an inconsistent model where we make impossible states possible...

magopian:
@n1k0 do you want to have a last look before I merge? I published it on https://mozilla-services.github.io/buildhub/

n1k0:
LGTM, great work

PR #7, opened 2017-05-05
[WIP] script to download and inspect archives (Fixes #202)

### LINKED ISSUE #202: System-addons information
We could have a collection of system-addons records. The question we want to answer is: what systemaddons (id + version) are shipped with a particular release and what updates users are supposed to run

We already had implemented the code:

* Download archive and inspect browser/modules content https://github.com/mozilla-services/buildhub/blob/5a6546bd368fcb7e566ebade9e1383b7c49583b8/jobs/inspect_archives.py
* Fetch update info from Balrog : https://github.com/mozilla-services/buildhub/blob/ca149b762faf2dc66385bb56698fcfae14648112/jobs/buildhub/sysaddons_updates.py

But there more that 300K records for Firefox builds, not all are relevant regarding system-addons since they appeared recently, but still, downloading 50M from archive and extracting them to inspect their content seems a bit airy-fairy.

We can get back on that after 1.0 with a finer strategy if it makes sense

leplatrem:
Fixes #202 

* [x] quick and dirty parallel download
* [ ] track highest processed timestamp
* [x] manage missing data 
* [ ] manage corrupt data
* [x] inspect system addons

### DIFF
--- a/jobs/inspect_archives.py
+++ b/jobs/inspect_archives.py
@@ -20,6 +20,7 @@ from kinto_http import cli_utils
 DEFAULT_SERVER = "https://kinto-ota.dev.mozaws.net/v1"
 DEFAULT_BUCKET = "build-hub"
 DEFAULT_COLLECTION = "archives"
+NB_RETRY_REQUEST = 5
 NB_WORKERS = 5  # CPU + 1
 
 
@@ -170,6 +171,7 @@ def main():
         description="Inspect and complete archives on Kinto",
         default_server=DEFAULT_SERVER,
         default_bucket=DEFAULT_BUCKET,
+        default_retry=NB_RETRY_REQUEST,
         default_collection=DEFAULT_COLLECTION)
 
     args = parser.parse_args(sys.argv[1:])
@@ -182,7 +184,6 @@ def main():
                 .format(**args.__dict__))
 
     client = cli_utils.create_client_from_args(args)
-    client.session.nb_retry = 1
 
     filters = {
         # "_since": "1493977727760",  # XXX: store previous timestamp in a file
--- a/jobs/inspect_archives.py
+++ b/jobs/inspect_archives.py
@@ -165,6 +165,8 @@ def main():
     args = parser.parse_args(sys.argv[1:])
 
     cli_utils.setup_logger(logger, args)
+    kinto_logger = logging.getLogger('kinto_http')
+    cli_utils.setup_logger(kinto_logger, args)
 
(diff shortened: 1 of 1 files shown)


jobs/inspect_archives.py
```diff
@@ -0,0 +1,221 @@
...
+            for chunk in stream.iter_content(chunk_size=1024):
+                if chunk:  # filter out keep-alive new chunks
+                    f.write(chunk)
+
+    return tempname
+
+
+def extract_application_metadata(ini_content):
+    config = configparser.ConfigParser()
+    config.read_string(ini_content.decode('ascii'))
```
Natim: Why ascii rather than utf-8?

jobs/inspect_archives.py
```diff
@@ -0,0 +1,221 @@
...
+        tar.close()
+    except:
+        logger.exception("Could not introspect archive %s" % url)
+
+    updated["systemaddons"] = systemaddons
+
+    has_changed = json.dumps(record, sort_keys=True) != json.dumps(updated, sort_keys=True)
+    if has_changed:
+        logger.info("Update metadata of %s" % updated)
+        # XXX why 412 here? why safe=False?
```
Natim: You will get a 412 if the updated record contains a last_modified and if it doesn't match the if-match header of the server.

leplatrem: Then something is wrong somewhere.

I'm sure the function is called once per record. We process one record at a time.

leplatrem:
Remaining things to do:

* [ ] Leave `systemaddons` attribute to `None` if installation has no `browser/features/` folder
* [ ] Unzip windows archives
* [ ] Set `channel` field

leplatrem:
Hmm, actually it would be more efficient to pull stuff from https://archive.mozilla.org/pub/firefox/candidates/51.0b6-candidates/build1/linux-x86_64/en-US/firefox-51.0b6.json

glasserc:
It seems like we are not going to pursue this approach, instead going with the JSON metadata files (#18, #19). Closing for now until we find a reason to use this technique instead.

leplatrem:
Unfortunately we'll have to keep this approach to:
* obtain the list (and versions) of built-in addons for recent versions
* obtain metadata for old versions (beyond [candidates folder](https://archive.mozilla.org/pub/firefox/candidates/)) and non english locales (json metadata is only there for en-US, see #19)

leplatrem:
This is never going to happen (I hope)

PR #11, opened 2017-05-11
Fetch updates from Balrog for every archive that has systemaddons.

leplatrem:
Another way of batching since exploration is not a tree :)


jobs/sysaddons_updates.py
```diff
@@ -0,0 +1,148 @@
...
+    channel = entry["source"]["tree"].split("/")[-1]
+
+    url = url.replace("{VERSION}", entry["target"]["version"])
+    url = url.replace("{BUILD_ID}", entry["build"]["id"])
+    url = url.replace("{BUILD_TARGET}", entry["target"]["platform"])
+    url = url.replace("{LOCALE}", entry["target"]["locale"])
+    url = url.replace("{CHANNEL}", channel)
+    url = url.replace("{OS_VERSION}", "default")
+    url = url.replace("{DISTRIBUTION}", "default")
+    url = url.replace("{DISTRIBUTION_VERSION}", "default")
```
magopian: Why not use `url.format(VERSION=entry["target"]["version"], BUILD_ID=...`?

jobs/sysaddons_updates.py
```diff
@@ -0,0 +1,148 @@
...
+
+        for i in range(nb_batches):
+            logger.debug("{}/{}".format(i, nb_batches))
+            records_subset = with_sysaddons[(i * AUS_BATCH_SIZE):((i + 1) * AUS_BATCH_SIZE)]
+
+            # Launch a bunch of parallel requests to AUS.
+            tasks = []
+            for record in records_subset:
+                coro = fetch_updates(session, AUS_URL, record)
+                task = asyncio.ensure_future(coro)
```
magopian: I've meant to ask in the previous PR but forgot: is it idiomatic to use `ensure_future` instead of a simple `await` or `yield from`? I would have personally used `await`

Issue #9, opened 2017-05-10
[ui] Add a page with code snippets


leplatrem:
In order to show the possibility of the underlying APIs, we could have a static page that:

* Shows a curl example to retrieve the revision of a buildid (for example)
* Shows kinto-http.(py|js) snippets to query the API (simple stuff like list of locales for a version) 
* Links to Kinto API docs
* Info to contact us (IRC, 

Thoughts?

n1k0:
Great idea!

glasserc:
Oops never mind

PR #12, opened 2017-05-11
Preliminary UI refactoring before #9

n1k0:
This is preliminary refactoring of the views before starting tackling #9. 

I removed the fixed sidebar thing, which was ugly and malfunctioning. I switched the many radio buttons to be just simple dropdowns, which I'm finding quite usable and gains us a lot of real estate.

Feedback? @magopian

### DIFF
--- a/ui/View.elm
+++ b/ui/View.elm
@@ -30,6 +30,11 @@ view model =
                             , filterSelector model.filterValues.platformList "Platforms" model.platformFilter (UpdateFilter << NewPlatformFilter)
                             , filterSelector model.filterValues.channelList "Channels" model.channelFilter (UpdateFilter << NewChannelFilter)
                             , filterSelector model.filterValues.localeList "Locales" model.localeFilter (UpdateFilter << NewLocaleFilter)
+                            , p [ class "text-right" ]
+                                [ button
+                                    [ class "btn btn-default", type_ "button", onClick (UpdateFilter ClearAll) ]
+                                    [ text "Clear all filters" ]
+                                ]
                             ]
                         ]
                     ]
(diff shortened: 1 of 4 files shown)


ui/View.elm
```diff
@@ -10,85 +10,77 @@ import Types exposing (..)
...
-                ]
+filterSelector : List String -> String -> String -> (String -> Msg) -> Html Msg
+filterSelector filters filterName checkedFilter onClickHandler =
+    let
+        optionView value =
+            option [] [ text value ]
+    in
+        div [ class "form-group" ]
+            [ label [] [ text filterName ]
+            , select [ class "form-control", onInput onClickHandler, value checkedFilter ] <|
```
magopian: is that format expected? Having the `<|` on the end of a line?

n1k0: that's elm-format autoformatting

n1k0:
I fixed a bug with the select widgets sometimes losing track of their selected value. I've also added a button to clear all filters at once, as well as rendering information about the number of matching builds and the list of filters applied:

ui/View.elm
```diff
@@ -10,85 +10,133 @@ import Types exposing (..)
...
-                    , onClick <| onClickHandler filterValue
-                    ]
-                    []
-                , text filterValue
+            []
+        ]
+
+
+filterSelector : List String -> String -> String -> (String -> Msg) -> Html Msg
+filterSelector filters filterName checkedFilter updateHandler =
```
magopian: Now that the filters are options in a select and not radiobuttons anymore, we might want to rename `checkedFilter` to `selectedFilter`?

n1k0:
I'm landing this.

n1k0:
I'm gonna address your latest comments directly on master, because I'm crazy.

PR #16, opened 2017-05-15
Merge filterInfos in the filter sidebar

magopian:
Trying something a bit different for the display of "filters information":

screen shot 2017-05-15 at 11 27 52

The idea is to replace the select boxes with the "filter info" label, to avoid having a select box with only the "all" and the selected filter options, making it (maybe?) a bit more ergonomic.

@n1k0 what do you think?

### DIFF
--- a/ui/View.elm
+++ b/ui/View.elm
@@ -176,13 +176,20 @@ filterSelector filters filterTitle selectedFilter updateHandler =
                 (List.map optionView ("all" :: filters))
 
         selectedFilterView =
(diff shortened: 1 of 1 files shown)


magopian:
Thanks for the feedback @n1k0 , what do you think of the following?

screen shot 2017-05-15 at 15 17 55

I'm still not very happy about the style of the `select`s, but it seems the only solution around that is to use "button dropdowns" and `ul`s, which i'm not sure is better (any thoughts?).

n1k0:
Yeah it's nicer, ship it! :)

PR #28, opened 2017-05-16
Experiment with bootstrap's button dropdown for better styling

magopian:
This is a very first draft. I'm not that happy about loosing the semantics of the `select` in favor of a plain `button`, and the addition of both jquery and `dropdown.js`. The upside is that the styling is possible, while it's not really with `select`s.

Even though the styling of buttons is easier, it seems i'll have to style them myself (and not use bootstrap's default css) because the result is pretty crappy:

screen shot 2017-05-16 at 14 24 06

So i'm wondering if i should just ditch this PR (and keep the `select`s), or go one step further, and implement the "dropdowns as buttons and ULs" myself (in elm) and style them...

@n1k0 any tips/thoughts?

### DIFF
--- a/ui/View.elm
+++ b/ui/View.elm
@@ -165,15 +165,17 @@ filterSelector : List String -> String -> String -> (String -> Msg) -> Html Msg
 filterSelector filters filterTitle selectedFilter updateHandler =
     let
         optionView value_ =
-            option [ value value_, selected (value_ == selectedFilter) ] [ text value_ ]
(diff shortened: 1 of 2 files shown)


n1k0:
I'm really afraid using an external lib that may mutate the dom and desync Elm's v-dom representation. Is this safe? Also I'm pretty much happy with the current version but that may just be me.

magopian:
Not following through with this idea, doesn't seem like a good one.

Issue #20, opened 2017-05-15
Deterministic records ids


leplatrem:
Currently we let the server choose the records ids (using POST).

@Natim suggested that we could have a relation between a release and its record.

Idea: We could use the download url (which is unique) to compute a uuid for example.

Natim:
My understanding after discussions is that we can use the following format to generate an unique ID for an archive.

```{product_name}/{channel}/{version}/{platform}/{locale}```

e.g: firefox/release/53.0b9/linux-i686/en-US

I would recommand using MD5 to generate a unique ID from it.

```python
>>> from uuid import UUID
>>> from hashlib import md5
>>> UUID(bytes=md5(b"firefox/release/53.0b9/linux-i686/en-US").digest())
UUID('50b4c8af-0cc3-fdea-ba3b-21e70d8e2404')
```

Natim:
Question from @glasserc:

Why not use directly: firefox_release_53.0b9_linux-i686_en-US

leplatrem:
I'm not sure if `channel` should be part of it

leplatrem:
And what about two consecutive nightlies? (same version, same channel)

Natim:
Today channel is part of the URL.
What is the URL for nightly version?

Natim:
And for mobile:

``fennec_releases_39.0b5_android-api-9_sl``

Natim:
Here is a gist of how it looks like: https://gist.github.com/Natim/049361533ccb861c2d770835ea188575

PR #30, opened 2017-05-16
Use a deterministic ID to create the records.

Natim:
Refs #20


jobs/scrape_archives.py
```diff
@@ -26,10 +27,17 @@
 logger = logging.getLogger(__name__)
 
 
+def build_record_id(record):
+    # Generate an ID that looks like that: firefox_release_53.0b9_linux-i686_en-US
+    return dirname(
+        record['download']['url'].replace(ARCHIVE_URL, '')).replace('/', '_').replace('.', '-')
```
leplatrem: nit: please use intermediary variables

leplatrem:
what was wrong with `ui` and `jobs` ?

data_tools/buildhub/utils.py
```diff
@@ -0,0 +1,22 @@
+from os.path import basename
+
+
+def build_record_id(url):
+    # Generate an ID that looks like that: firefox_53.0b9_linux-i686_en-US
+    parts = basename(url).split('.')
+    url_parts = url.split('/')
+
+    product_name, major_version = parts[0].split('-')
```
leplatrem: Maybe we could pass an archive record instead of parsing this again. (plus, `mobile` != `fennec`)

data_tools/buildhub/utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+    if 'nightly' in url:
```
leplatrem: Can you add an example of URI here?

data_tools/buildhub/utils.py
```diff
@@ -0,0 +1,22 @@
...
+        build_date = '-'.join(url_parts[8].split('-')[:6])
+        version_number = '{}.{}'.format(major_version, parts[1])
+        version = '{}_{}'.format(build_date, version_number)
+        locale = parts[2]
+        arch = parts[3]
```
leplatrem: See previous comment about passing a record instead of parsing this again

data_tools/setup.py
```diff
@@ -0,0 +1,57 @@
...
+
+README = read_file('README.rst')
+CHANGELOG = read_file('CHANGELOG.rst')
+CONTRIBUTORS = read_file('CONTRIBUTORS.rst')
+
+REQUIREMENTS = [
+    "aiohttp",
+    "backoff",
+    "kinto-http",
+    "MozillaPulse",
```
leplatrem: This one is not compatible with Python3

jobs/tests.py
```diff
@@ -1,12 +0,0 @@
...
-from main import url2version
-
-class TestUrl2Version(unittest.TestCase):
-
-    def test_extract_from_url(self):
-        assert url2version("/firefox-55.0a1.en-US.win32.zip") == "55.0a1"
-
-
-if __name__ == '__main__':
-    unittest.main()
```
leplatrem: Where did this go?

data_tools/buildhub/jobs/scrape_archives.py
```diff
@@ -29,7 +30,8 @@
 def publish_records(client, records):
     with client.batch() as batch:
         for record in records:
-            batch.create_record(record)
+            record_id = build_record_id(record['download']['url'])
```
leplatrem: I would do this in the `archive()` function instead since the `id` field can be put in the `data` mapping.

Natim:
We can probably keep `ui` but when we talked about it with Mago it feld more explicit to talk about `web_ui`.

About the python scripts, what we wanted to do is a python package that we could import in our tests and manage version about.

data_tools/buildhub/utils.py
Natim: We have test that cover this. Here we are taking the name in the archive name not the first parts of the URL.

Natim:
> I'm not super comfortable with this since we have other PRs that will conflict.

Don't worry too much about it because most of the file where just moved around not changed and git handle file rename really efficiently so it shouldn't be a burden at all.

jobs/listen_pulse.py
```diff
@@ -1,142 +0,0 @@
...
-        except KeyboardInterrupt:
-            raise
-        except IOError:
-            # Sometimes there is a socket timeout. Just call listen again; this
-            # is fairly common and not worth logging.
-            pass
-        except:
-            # It is possible for rabbitmq to throw other exceptions. Just log
-            # them and move on.
-            traceback.print_exc()
```
leplatrem: Did you delete `listen_pulse`?

data_tools/buildhub/jobs/scrape_archives.py
```diff
@@ -159,7 +151,7 @@ def archive_url(product, version=None, platform=None, locale=None, nightly=None)
 async def fetch_nightlies(session, queue, product, client):
     # Check latest known version on server.
     filters = {
-        "source.product": product,
+        "target.product": product,
```
leplatrem: Why this change here?

data_tools/buildhub/utils.py
```diff
@@ -0,0 +1,93 @@
...
+    url_parts = url.split('/')
+
+    product_name, major_version = parts[0].split('-')
+    channel = 'release'
+
+    if 'nightly' in url:
+        version = '{}.{}'.format(major_version, parts[1])
+        locale = parts[2]
+        platform = parts[3]
+        if 'aurora' in url:
```
leplatrem: When does this happen?

data_tools/buildhub/utils.py
```diff
@@ -0,0 +1,93 @@
...
+    if record["target"]["channel"] == 'nightly':
+        url_parts = record["download"]["url"].split('/')
+        date_parts = url_parts[8].split('-')
+        date = '-'.join(date_parts[:6])
+        version = '{}_{}'.format(date, version)
+    return '{target[product]}_{version}_{target[platform]}_{target[locale]}'.format(
+        target=record['target'], version=version).replace('.', '-').lower()
+
+
+def build_record_id_from_url(url):
```
leplatrem: I can't find any reference of this function, where is this used?

data_tools/buildhub/utils.py
```diff
@@ -0,0 +1,93 @@
+from os.path import basename
+
+
+def build_record_from_url(url):
+    record = {}
+    parts = basename(url).split('.')
+    url_parts = url.split('/')
+
+    product_name, major_version = parts[0].split('-')
```
leplatrem: I'm not super confortable with the fact that parse the product from url, since the product is something we have in the code in the `PRODUCTS` tuple.

data_tools/buildhub/utils.py
```diff
@@ -0,0 +1,93 @@
...
+
+def build_record_id(record):
+    version = record["target"]["version"]
+    if record["target"]["channel"] == 'nightly':
+        url_parts = record["download"]["url"].split('/')
+        date_parts = url_parts[8].split('-')
+        date = '-'.join(date_parts[:6])
+        version = '{}_{}'.format(date, version)
+    return '{target[product]}_{version}_{target[platform]}_{target[locale]}'.format(
+        target=record['target'], version=version).replace('.', '-').lower()
```
leplatrem:
```py
id_ = '{product}_{version}_{platform}_{locale}'.format(**record["target"], version=version)
return id_.replace('.', '-').lower()
```

data_tools/buildhub/utils.py
Natim: This was a step to move from the previous code to the rework that you asked. If we don't use it we can remove it. Right now it is used to make sure that the two behaves the same way.

jobs/listen_pulse.py
Natim: Yes, because it is not compatible with Python3 and right now you said you don't want to rely on poling for changes.

data_tools/buildhub/jobs/scrape_archives.py
Natim: Because after multiple discussion we said that it didn't made sense to have it separated from target.

data_tools/buildhub/utils.py
Natim: The products tuple doesn't contains fennec, so I cannot use it to make my test to pass.

We decided together that the ID should be generated from the URL and the URL only.
Then you asked that it comes from the record, but the way the record is currently generated doesn't provide the information to build the ID from it.

You want me to iterate but in the meantime you are the one asking for those refactoring because without them the tests don't pass.

I would be glad to pair review this so that we can discuss our views on the project and find a consensus.

data_tools/buildhub/utils.py
Natim: It happens for the Aurora URLs (there is a test covering it.

https://github.com/mozilla-services/buildhub/pull/30/commits/63918d49b6035c7b2af3071571d9ea15b55d280d

data_tools/buildhub/utils.py
Natim: The goal I had in mind with this PR was simply to generate the ID from the URL.
You asked me in a previous review to use the record rather the URL, but the current record doesn't contains the mandatory information

data_tools/buildhub/utils.py
Natim: `TypeError: format() got multiple values for keyword argument 'version'`

Natim:
About the deletion of pulse I would be glad to know how you want to support some code that is Python3 only (asyncio) with some code that is Python2 only (MozillaPulse) in the same codebase.

Natim:
The current scrape_archive script is quite broken (I blame the fact that it has no tests).

With the fact that we don't rely on the URL anymore, we start to see failing IDs such as:

- `firefox_2017-05-01-10-01-39_55-0a1_linux-x86_64-tar_zh-tw` (notice the `tar` in the platform)

leplatrem:
>  I would be glad to know how you want to support some code that is Python3 only (asyncio) with some code that is Python2 only (MozillaPulse) in the same codebase.

You were the one to choose to make this a full package :) It was just a set of scripts! We could have waited for #39 to make this repo a proper package for example

leplatrem:
> The current scrape_archive script is quite broken (I blame the fact that it has no tests).

Yes, it well could be! please open individual issues. 

> Fixing in #45

This PR is called adding tests, but also covers packaging. It will then also cover this platform if I understand correctly! 
We should meet and pair on merging all this mess and go back to small issue/changesets, because this won't be playable (I'm too old for this :D)

Natim:
> You were the one to choose to make this a full package :) It was just a set of scripts!

The problem I wanted to solve is the fact that none of this scripts are tested, which makes them really hard to refactor and even harder to maintain.

For me creating a package was a first step to be able to add tests for the deterministic ID feature.
I feel so bad reading your review... :(

Issue #29, opened 2017-05-16
Clarify and fix scripts for repository/tree


leplatrem:
Currently we don't have clear semantics about tree/channel etc.

I propose that we keep:
* `repository`: full repository URL (eg. `https://hg.mozilla.org/releases/mozilla-beta`)
* ~~`channel`: update channel (for AUS, `beta`, `nightly`, `release`)~~
* `tree`: path of repository relative to root (eg. `releases/mozilla-beta`)→ useful?

Related #17

PR #33, opened 2017-05-16
First draft of a JSON schema.

Natim:
- [x] Add a documentation
- [ ] Add a record example (complete and uncomplete) when we are uncomplete, when we are complete etc.


schema/archive.json
```diff
@@ -0,0 +1,210 @@
+{
+  "collections": {
+    "archives": {
```
leplatrem: Why this level of embedded JSON?

schema/archive.json
```diff
@@ -0,0 +1,210 @@
...
+      "name": "archives",
+      "config": {
+        "displayFields": [
+          "id", "target.product", "target.channel", "target.version",
+          "target.platform", "target.locale"
+        ],
+        "schema": {
+          "title": "Archives",
+          "description": "Mozilla software builds.",
```
leplatrem: Let's replace builds with «product releases» here. While waiting for #35

schema/archive.json
```diff
@@ -0,0 +1,210 @@
...
+            "id": {
+              "type": "string",
+              "title": "ID",
+              "description": "Build Hub ID"
+            },
+            "schema": {
+              "type": "integer",
+              "title": "Schema version",
+              "description": "Schema version"
+            },
```
leplatrem: Are those two mandatory? Why not `last_modified`?

schema/archive.json
```diff
@@ -0,0 +1,210 @@
...
+              "properties": {
+                "build_id": {
+                  "type": "string",
+                  "title": "Build ID",
+                  "description": "Build ID"
+                },
+                "build_date": {
+                  "type": "string",
+                  "title": "Build date",
+                  "description": "i.e: 2017-05-11",
```
leplatrem: ISO format

schema/archive.json
```diff
@@ -0,0 +1,210 @@
...
+                  "pattern": "^(19|2[0-5])[0-9]{2}-((0[0-9])|(1[0-2]))-(([0-2][0-9])|(3[0-1]))$"
+                },
+                "sunset_date": {
+                  "type": "string",
+                  "title": "Sunset date",
+                  "description": "i.e: 2017-08-24",
+                  "pattern": "^(19|2[0-5])[0-9]{2}-((0[0-9])|(1[0-2]))-(([0-2][0-9])|(3[0-1]))$"
+                },
```
leplatrem: Is this new? Maybe we should keep the schema aligned with what we currently have in the scripts, and add an issue to add this new field

schema/archive.json
```diff
@@ -0,0 +1,210 @@
...
+                "build_type": {
+                  "type": "string",
+                  "title": "Build type",
+                  "description": "Build type",
+                  "enum": [
+                    "nightly",
+                    "beta",
+                    "aurora",
+                    "release",
```
leplatrem: Build type is always `opt` as far as I know. We'll probably remove it...

schema/archive.json
```diff
@@ -0,0 +1,210 @@
...
+            "schema": {
+              "type": "integer",
+              "title": "Schema version",
+              "description": "Schema version"
+            },
+            "build": {
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "build_id": {
```
leplatrem: field is `id`

schema/archive.json
```diff
@@ -0,0 +1,210 @@
...
+                  "type": "string",
+                  "title": "Build ID",
+                  "description": "Build ID"
+                },
+                "build_date": {
```
leplatrem: field is `date`

schema/archive.json
```diff
@@ -0,0 +1,210 @@
...
+                  "title": "Featured version?"
+                },
+                "has_builds": {
+                  "type": "boolean",
+                  "title": "Has builds?"
+                },
+                "is_rapid_beta": {
+                  "type": "boolean",
+                  "title": "Is rapid beta?"
+                }
```
leplatrem: Ditto about new fields: there are not here yet.

schema/archive.json
```diff
@@ -0,0 +1,210 @@
...
+                "url": {
+                  "type": "string",
+                  "format": "url",
+                  "title": "URL",
+                  "description": "URL of the archive"
+                }
+              }
+            }
+          }
+        },
```
leplatrem: System addons is missing here I think

schema/archive.json
```diff
@@ -0,0 +1,210 @@
...
+              "properties": {
+                "tree": {
+                  "type": "string",
+                  "title": "Tree",
+                  "description": "i.e mozilla-central"
+                },
+                "revision": {
+                  "type": "string",
+                  "title": "Revision number in the tree"
+                }
```
leplatrem: Maybe add `repository` (see #29)

schema/archive.json
leplatrem: Maybe add a Readme in the folder to show how to use this

schema/archive.json
```diff
@@ -0,0 +1,210 @@
...
+                "beta_number": {
+                  "type": "integer",
+                  "title": "Beta number",
+                  "description": "Beta number (i.e: 7)"
+                },
+                "version_sort": {
+                  "type": "string",
+                  "title": "A value to sort records by version number",
+                  "description": "A value to sort records by version number. (i.e: 052001001r000 (release), 054000000a002 (nightly), 054000000b007 (beta), 045009000x000 (esr))"
+                },
```
glasserc: Is this also a new field?

schema/archive.json
glasserc: I just realized that I didn't understand why `product` is part of `source`, and I was going to ask here, but I see that it isn't here, I guess also because this schema is aspirational?

mostlygeek:
Where/what will this schema be used for?

glasserc:
I think it's meant to:

- Document the fields available in the `archives` collection, i.e. serve as an interface contract.
- Enforce that records conform to a known schema (and don't e.g. randomly drop important fields).

mostlygeek:
A schema means we need to do some normalization / transformation somewhere. I think schema enforcement should be in an application layer. Seems we should discuss that first before we work on normalizing data.

glasserc:
Wait, what? Isn't the entire project about normalization/transformation? Isn't that the sole and foremost responsibility of the scrapers? What do you mean by "application layer"? Is there a reason not to also have enforcement at the database layer?

glasserc:
Wait, I think I understand what happened. This PR adds a [JSON Schema](http://json-schema.org/) document describing the format of records in Kinto (which are JSON records). This doesn't represent any kind of message/request/response format.

schema/archive.json
Natim: This is just in case we need to handle multiple collection schema in that file as it was the case for the AMO blocklists. Yes a README would be useful.

schema/archive.json
Natim: I know for sure that last_modified is stripped by the admin. Probably these two are too.

schema/archive.json
Natim: So the plan is to build a schema that meet our expectation rather than trying to fit the current schema.

1. We build a proper schema
2. We upgrade the scripts to fit our schema.

The new fields are the one missing from our schema that are part of the product_version table in socorro.

schema/archive.json
Natim: What does opt means? If you look at the table that Will sent us, you will see that they are using the one from the enum list.

schema/archive.json
Natim: I am not sure if they are part of our current use cases. But that is something we need to discuss indeed.

leplatrem:
This is the kinto-elasticsearch schema:

```
build-hub:
  collections:
    releases:
      data:
        index:schema:
          properties:
            id:
              type: "keyword"
            last_modified:
              type: "long"
            build:
              properties:
                id:
                  type: "keyword"
                date:
                  type: "date"
                  format: "date"
            source:
              properties:
                product:
                  type: "keyword"
                repository:
                  type: "keyword"
                tree:
                  type: "keyword"
                revision:
                  type: "keyword"
            target:
              properties:
                platform:
                  type: "keyword"
                locale:
                  type: "keyword"
                version:
                  type: "keyword"
                channel:
                  type: "keyword"
            download:
              properties:
                url:
                  type: "keyword"
                mimetype:
                  type: "keyword"
                size:
                  type: "long"
                date:
                  type: "date"
```

leplatrem:
This current schema does not «support `null` or object». In the current state, this is required for the scrape_archives script to run properly.

But `null` is what we chose to specify that a field is not available. For example, `systemaddons = null` means systemaddons not available for this version versus `systemaddons = []` means the list of systemaddons is empty.

The diff here suggests that we should remove the fields when the information is not available, because we can't express none or object easily. This is not a minor change (cf. UI) and we should probably decide a clear strategy (add complexity to schema with anyof, change the scrape archives and UI in a separate PR, or whatever).

Natim:
Yes, I would be in favour of not adding properties to object if we don't know the value rather than setting it null.
I started the work on https://github.com/mozilla-services/buildhub/pull/56 to fix the UI with regards to the schema.

leplatrem:
> I started the work on #56 to fix the UI with regards to the schema.

Please rename the PR title then :) Especially if switch to releases collection requires introducing pagination.

jobs/initialization.yml
```diff
@@ -0,0 +1,205 @@
...
+                  type: "keyword"
+                mimetype:
+                  type: "keyword"
+                size:
+                  type: "long"
+                date:
+                  type: "date"
+        displayFields:
+          - id
+          - target.product
```
Natim: source.product

Issue #27, opened 2017-05-15
WebUI should create permalink URI for resources


mostlygeek:
The URI address in the browser should update when we change filters. This would make it easier for other web tools (mozreview, bugs, phrabricator, etc) to link back to build/release information.

mostlygeek:
closing this in favor of #40. No point in having discussions in two places.

PR #40, opened 2017-05-17
URLify the web UI

magopian:
Addresses #27 

- [x] URLify the "docs" and "builds" header buttons
- [x] URLify the filters
- [x] URLify specific builds

### DIFF
--- a/ui/Main.elm
+++ b/ui/Main.elm
@@ -68,8 +68,12 @@ update msg ({ filterValues } as model) =
 
                         NewBuildIdSearch value ->
                             { model | buildIdFilter = value }
+
+                updatedRoute =
+                    { model | route = routeFromFilters updatedFilters }
             in
-                updateModelWithFilters updatedFilters ! []
+                updateModelWithFilters updatedRoute
+                    ! [ newUrl <| urlFromRoute updatedRoute ]
 
         UrlChange location ->
-            updateModelWithFilters (parsePage model location) ! []
+            updateModelWithFilters (routeFromUrl model location) ! []
--- a/ui/Main.elm
+++ b/ui/Main.elm
@@ -1,15 +1,17 @@
 module Main exposing (..)
 
-import Html exposing (..)
 import Model exposing (..)
+import Navigation exposing (..)
 import Set
 import Types exposing (..)
+import Url exposing (..)
 import View exposing (..)
 
 
 main : Program Never Model Msg
 main =
-    Html.program
+    Navigation.program
+        UrlChange
         { init = init
         , view = view
         , update = update
@@ -20,9 +22,6 @@ main =
 update : Msg -> Model -> ( Model, Cmd Msg )
 update msg ({ filterValues } as model) =
     case msg of
-        ChangeView view ->
-            { model | currentView = view } ! []
-
         BuildRecordsFetched (Ok buildRecordList) ->
             { model
                 | builds = buildRecordList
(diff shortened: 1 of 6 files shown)


mostlygeek:
Currently the URLs use the uuid. I think it's worthwhile to make the url something more predictable: 

Perhaps modeling after the AUS (balrog/updater service) urls is a good start: 

`https://aus5.mozilla.org/update/6/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%SYSTEM_CAPABILITIES%/%DISTRIBUTION%/%DISTRIBUTION_VERSION%/update.xml`

Natim:
> Currently the URLs use the uuid.

It won't be the case anymore now that #30 landed.

magopian:
Regarding https://github.com/mozilla-services/buildhub/pull/40#issuecomment-302804086 do we want to keep the same url structure? And thus the same "catalog categories"?

The question is: for the "catalog-ifying" of the UI (selecting a few things before displaying a huge list of builds, with a huge list of filters), what do we want to pre-select, and in which order?

Eg:
1/ select the product
2/ select the version
3/ select the platform

Or
1/ select the product
2/ select the channel
3/ select the version

Or a mix of those? Also, do we allow/want to select "all" for any of those?

Natim:
What about a selection of the product and then a selection of which kind of channels, versions or platforms you want to filter on.

A bit like searching for bugs in Bugzilla

example2

magopian:
I believe this is now ready for review: changing the url will update the filters, changing the filters will update the url. I've created two new issues to follow up on that:
- #64 Remove the "trees" filter which is redundant with the "channels" filter
- #65 Only use the `Route`, and drop the redundant `*Filter`s from the model

ui/Main.elm
```diff
@@ -75,93 +69,11 @@ update msg ({ filterValues } as model) =
                         NewBuildIdSearch value ->
                             { model | buildIdFilter = value }
 
-                updatedFilteredBuilds =
-                    applyFilters updatedModelWithFilters
+                updatedRoute =
```
n1k0: Nit: this is a `Model` value, `updatedRoute` sounds strange. `updatedModel` maybe?

ui/Url.elm
```diff
@@ -0,0 +1,204 @@
...
+                    , channelFilter = "all"
+                    , platformFilter = "all"
+                    , versionFilter = "all"
+                    , localeFilter = "all"
+                }
+
+
+urlFromRoute : Model -> String
+urlFromRoute model =
+    case model.route of
```
n1k0: Nit: If we're only using the `route` field, it sounds faire to just accept a `Route` argument, not a full `Model` one.

ui/Url.elm
```diff
@@ -0,0 +1,204 @@
...
+                ++ locale
+
+        _ ->
+            "#/builds/"
+
+
+routeFromFilters : Model -> Route
+routeFromFilters model =
+    if model.buildIdFilter /= "" then
+        BuildIdView model.buildIdFilter
```
n1k0: Nit: How about destructuring to decrease verbosity here?

```elm
routeFromFilters model =
    let
        { buildIdFilter, productFilter, channelFilter, platformFilter, versionFilter, localeFilter }
            = model
    in
        if buildIdFilter /= "" then
            BuildIdView buildIdFilter
        else if localeFilter /= "all" then
            LocaleView productFilter channelFilter platformFilter versionFilter localeFilter
```

Also it really feels like filters should be a `Dict`, but let's not waste time with this yet.

magopian: I filed #65 already to remove them ;)

magopian:
@n1k0 I believe I've addressed all your comments, thanks for the review! If you're happy with the changes, please let me know and I'll merge them.

n1k0:
Review has been approved already, ship it! :)

Le 30 mai 2017 2:57 PM, "Mathieu Agopian" <notifications@github.com> a
écrit :

Issue #39, opened 2017-05-17
Migrate MozillaPulse client to Python 3


leplatrem:
If we want to leverage what we receive in Pulse, we'll have to migrate it to Python 3

Rework first version of patch here: https://bugzilla.mozilla.org/show_bug.cgi?id=1294230

mostlygeek:
A bigger question: should we standardize on Python3? In #41 I want to deploy aggregator code in a docker container. We'll need to pick a python base container.

Natim:
We are relying for the scrapping scripts on asyncio which is a event loop API that was brought to Python 3 so yes, we should :)

leplatrem:
> A bigger question: should we standardize on Python3?

Yes. Standardizing on Python2 would be like targetting Windows XP

leplatrem:
Port of MozillaPulse to Python3 was merged. Waiting for release on Pypi.

leplatrem:
Was moved to Github https://github.com/mozilla-services/mozillapulse
And version 1.3 was released with Py3 support :)

PR #45, opened 2017-05-18
Add scraping tests

Natim:
Based on #30


glasserc:
Needs a rebase now that I updated kinto-http.py to 8.0.1..

Nice use of `parametrize`.

What happened to `requirements.txt` ? I think we still need it.

Personally I think we ought to keep the `listen_pulse` script, even if we can't use it yet (#39) . But I'll let you and @leplatrem fight that out.

I'm not crazy about the name `data_tools`, how about `importers`?

leplatrem:
What is the status of this PR? Is it supposed to replace #30? Why is there a commit titled `Use a deterministic ID to create the records.` then?

> Personally I think we ought to keep the listen_pulse script, even if we can't use it yet (#39) . But I'll let you and @leplatrem fight that out.

So do I :)

> I'm not crazy about the name data_tools, how about importers?

I still wonder what was wrong with `ui` and `jobs`, or at least why was it necessary to change that now and make sure we «bikeshed early, bikeshed often»

data_tools/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,7 @@
+Contributors
+============
+
+* Mathieu Agopian <magopian@mozilla.com>
+* Rémy Hubscher <rhubscher@mozilla.com>
+* Mathieu Leplatre <mathieu@mozilla.com>
+* Nicolas Perriault <nperriault@mozilla.com>
```
leplatrem: Ethan missing :)

data_tools/CHANGELOG.rst
```diff
@@ -0,0 +1,12 @@
...
+
+This document describes changes between each past release.
+
+0.1 (unreleased)
+----------------
+
+**Initial version**
+
+- Scraping tools from archive
+- Scraping tools for systemaddons
```
leplatrem: Technically the systemaddons thing is not a scraper. So just write something short like:
* Scrape product releases and nightlies from https://archives.mozilla.org
* Fetch system addons update information from Balrog

data_tools/buildhub/utils.py
```diff
@@ -0,0 +1,62 @@
+import re
+FILE_EXTENSIONS = "zip|tar.gz|tar.bz2|dmg|apk"
```
leplatrem: missing blank line

data_tools/buildhub/utils.py
```diff
@@ -0,0 +1,62 @@
...
+
+def build_record_id(record):
+    version = record["target"]["version"]
+    if record["target"]["channel"] == 'nightly':
+        url_parts = record["download"]["url"].split('/')
+        date_parts = url_parts[8].split('-')
+        date = '-'.join(date_parts[:6])
+        version = '{}_{}'.format(date, version)
+    return '{source[product]}_{version}_{target[platform]}_{target[locale]}'.format(
+        **record, version=version).replace('.', '-').lower()
```
leplatrem: This looks over complex (see comment in previous PR)

data_tools/buildhub/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    - firefox-55.0a1.ach.win64.zip
+    - firefox-55.0a1.bn-IN.mac.dmg
+
+    And things we'll want to ignore:
+
+    - firefox-55.0a1.en-US.linux-i686.talos.tests.zip
+    - firefox-55.0a1.en-US.mac.crashreporter-symbols.zip
+    """
+    re_nightly = re.compile(r"\w+-(\d+.+)\.([a-z]+(\-[A-Z]+)?)\.(.+)\.({})$".format(
+        FILE_EXTENSIONS))
```
leplatrem: nit: this line breaking is not elegant. How about splitting the regex in two parts?

data_tools/tests/test_utils.py
```diff
@@ -0,0 +1,252 @@
...
+            "url": "https://archive.mozilla.org/pub/mobile/releases/42.0b2/android-api-9/fr/"
+            "fennec-42.0b2.fr.android-arm.apk",
+        }
+    },
+]
+
+
+# Build record_id from record
+@pytest.mark.parametrize("record", RECORDS)
+def test_build_record_id_from_url(record):
```
leplatrem: The comment and the function name don't match. Remove (useless?) comment?

data_tools/tests/test_utils.py
```diff
@@ -0,0 +1,252 @@
...
+
+WRONG_RELEASE_FILENAMES = [
+    ("firefox", "firefox-1.5.0.5.tar.gz.asc"),
+    ("firefox", "firefox-52.0.win32.sdk.zip"),
+]
+
+
+@pytest.mark.parametrize("product,filename", WRONG_RELEASE_FILENAMES)
+def test_wrong_release_filename(product, filename):
+    assert not is_release_filename(product, filename)
```
leplatrem: Nice tests 👍 👍

Natim:
> What is the status of this PR? Is it supposed to replace #30?

It is not, we should merge #30 first.

Because you didn't want me to keep working on things, I created this PR to be able to move forward while we are reviewing #30.

Natim:
> Personally I think we ought to keep the listen_pulse script, even if we can't use it yet

I don't agree, the code is accessible on Github and the plan is to port it to Python 3 as soon as we merge #30.

Natim:
> I still wonder what was wrong with ui and jobs,

jobs is to restrictive for what the buildhub python package is meant to be.
The names are things that we brainstormed with @magopian while being in Paris. We find them more explicit and self documented than `ui` and `jobs` but I totally don't care about naming so if you feel that it doesn't add any value, we can keep the old names.

PR #62, opened 2017-05-29
Remove null values.

Natim:
This is a first step in order to build a Kinto compatible JSONSchema.

### DIFF
--- a/jobs/buildhub/scrape_archives.py
+++ b/jobs/buildhub/scrape_archives.py
@@ -91,34 +91,10 @@ def latest_known_version(client, product):
 
 
 def archive(product, version, platform, locale, url, size, date, metadata=None):
-    build = None
-    revision = None
-    repository = None
-    tree = None
-
     channel = guess_channel(url, version)
 
-    if metadata:
-        # Example of metadata:
-        #  https://archive.mozilla.org/pub/thunderbird/candidates \
-        #  /50.0b1-candidates/build2/linux-i686/en-US/thunderbird-50.0b1.json
-        revision = metadata["moz_source_stamp"]
-        channel = metadata.get("moz_update_channel", channel)
-        repository = metadata["moz_source_repo"].replace("MOZ_SOURCE_REPO=", "")
-        tree = repository.split("hg.mozilla.org/", 1)[-1]
-        buildid = metadata["buildid"]
-        builddate = datetime.datetime.strptime(buildid[:12], "%Y%m%d%H%M").isoformat()
-        build = {
-            "id": buildid,
-            "date": builddate,
-        }
-
     record = {
-        "build": build,
         "source": {
-            "revision": revision,
-            "repository": repository,
-            "tree": tree,
             "product": product,
         },
         "target": {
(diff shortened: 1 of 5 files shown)


jobs/buildhub/scrape_archives.py
```diff
@@ -112,9 +89,25 @@ def archive(product, version, platform, locale, url, size, date, metadata=None):
...
+        channel = metadata.get("moz_update_channel", channel)
+        repository = metadata["moz_source_repo"].replace("MOZ_SOURCE_REPO=", "")
+        record['source']['repository'] = repository
+        record['source']['tree'] = repository.split("hg.mozilla.org/", 1)[-1]
+        buildid = metadata["buildid"]
+        builddate = datetime.datetime.strptime(buildid[:12], "%Y%m%d%H%M").isoformat()
+        record['build'] = {
+            "id": buildid,
+            "date": builddate,
+        }
```
leplatrem: This is a micronit, but could you group together the obtention of metadata values and appart the assignations into `record`?

ui/Decoder.elm
```diff
@@ -28,26 +28,27 @@ buildDecoder =
...
 
 sourceDecoder : Decoder Source
 sourceDecoder =
     decode Source
         |> required "product" string
-        |> required "tree" string
-        |> required "revision" (nullable string)
+        |> optional "tree" (nullable string) Nothing
+        |> optional "revision" (nullable string) Nothing
+        |> optional "repository" (nullable string) Nothing
```
leplatrem: Shouldn't be `string Nothing` only? Remove `nullable`?

Natim:
> Did leave nullable in Elm types for a particular reason?

Yes this was to be able to decode to a Maybe and use ``Nothing`` as the default value in the decoder.
So that if the field is not present we default its value to null in the Elm type.

ui/Decoder.elm
magopian: The elm-decode-pipeline needs the decoder to be of the same type as the fallback, so a `Maybe` in both case. And the only way (that I know of) to do that is to use `nullable`.

ui/Model.elm
```diff
@@ -118,11 +118,11 @@ applyFilters model =
         |> List.filter
             (\buildRecord ->
                 (recordStringEquals (.source >> .product) model.productFilter) buildRecord
-                    && (recordMaybeStringEquals (.target >> .version) model.versionFilter) buildRecord
+                    && (recordStringEquals (.target >> .version) model.versionFilter) buildRecord
                     && (recordStringEquals (.target >> .platform) model.platformFilter) buildRecord
-                    && (recordMaybeStringEquals (.target >> .channel) model.channelFilter) buildRecord
```
magopian: I believe `recordMaybeStringEquals` isn't used anywhere else, so you can remove its code entirely.

ui/View.elm
```diff
@@ -314,7 +325,7 @@ viewSourceDetails source =
                 Just revision ->
                     let
                         mozillaHgUrl =
-                            treeToMozillaHgUrl source.tree
+                            treeToMozillaHgUrl <| Maybe.withDefault "" source.tree
```
magopian: Would it make sense to change the signature of `treeToMozillaHgUrl` to take a `Maybe String` instead?

ui/View.elm
```diff
@@ -337,39 +348,35 @@ viewSourceDetails source =
             , tbody []
                 [ tr []
                     [ td [] [ text source.product ]
-                    , td [] [ text source.tree ]
+                    , td [] [ text <| Maybe.withDefault "" source.tree ]
```
magopian: for the `record.build` above you use "unknown" as the default, do you want to use the same here?

magopian:
Thanks for this @Natim <3

PR #63, opened 2017-05-29
Paginate build list.

n1k0:
Note: this is not intended for landing as is. This patch illustrates how the new elm-kinto pagination system works, with a working implementation.

We'll definitely wait for #40 to land first before landing this, especially as this can't fully work with in-memory filtering.

### DIFF
--- a/ui/Main.elm
+++ b/ui/Main.elm
@@ -23,6 +23,25 @@ update : Msg -> Model -> ( Model, Cmd Msg )
 update msg ({ filterValues } as model) =
     case msg of
         BuildRecordsFetched (Ok buildsPager) ->
+            updateModelWithFilters
+                { model
+                    | buildsPager = buildsPager
+                    , filteredBuilds = buildsPager.objects
+                    , loading = False
+                }
+                ! []
+
+        BuildRecordsFetched (Err err) ->
+            let
+                _ =
+                    Debug.log "An error occured while fetching the build records" err
+            in
+                model ! []
+
+        LoadNextPage ->
+            model ! [ getNextBuilds model.buildsPager ]
+
+        BuildRecordsNextPageFetched (Ok buildsPager) ->
             let
                 newPager =
                     Kinto.updatePager buildsPager model.buildsPager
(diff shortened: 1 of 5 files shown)


ui/Model.elm
```diff
@@ -27,10 +27,21 @@ getBuildRecordList : Cmd Msg
 getBuildRecordList =
     client
         |> Kinto.getList recordResource
+        |> Kinto.limit 10
```
magopian: should we put something like 100 here? It seems filtering 100 objects in memory (if we'll keep doing that?) shouldn't be too memory/cpu intensive

n1k0: :+1:

magopian:
I believe this is now read to review, @n1k0 @Natim ?

ui/Main.elm
```diff
@@ -32,6 +38,29 @@ update msg ({ filterValues } as model) =
...
+                        | buildsPager = newPager
+                        , filteredBuilds = newPager.objects
+                        , loading = False
+                    }
+                    ! []
+
+        BuildRecordsNextPageFetched (Err err) ->
+            let
+                _ =
+                    Debug.log "An error occured while fetching the next page of build records" err
```
n1k0: Nit: We should file an issue to have friendly error messages in the UI.

ui/Main.elm
```diff
@@ -21,8 +22,13 @@ main =
 update : Msg -> Model -> ( Model, Cmd Msg )
 update msg ({ filterValues } as model) =
     case msg of
-        BuildRecordsFetched (Ok buildRecordList) ->
-            updateModelWithFilters ({ model | builds = buildRecordList, loading = False })
+        BuildRecordsFetched (Ok buildsPager) ->
+            updateModelWithFilters
+                { model
+                    | buildsPager = buildsPager
+                    , filteredBuilds = buildsPager.objects
```
n1k0: I don't think this is actually used anymore, and should be removed.

magopian: Very well spotted, thanks!
And pulling that string, it came with a bunch of other dead code <3

ui/Main.elm
magopian: :+1: #79

Issue #41, opened 2017-05-17
Operational Spec for BuildHub Version 1.0


mostlygeek:
This issue is to discuss how we want to structure the project to make it easy to configure and deploy for the operations team. Currently there are are server side and client side components. The server side components are daemons and cron job compatible scrapers. The client side is a static HTML website. 

Data is stored in a kinto instance which is backed by Postgresql RDS. This is relatively slow moving and shouldn't require much operational work other than standing it up and keeping it updated. 

Server side code
============

* ship a container with all server based code
* follow [dockerflow](https://github.com/mozilla-services/dockerflow)
* Build/test container in circle.ci
* Publish container to hub.docker.com/mozilla/buildhub
* A single container with all of the scripts / workers / scrapers / etc
* version.json written into the container, accessible by `run.sh version` 
* a `run.sh` file as the container's `ENTRYPOINT`. 

run.sh
-------------
Buildhub's docker container is a toolbox and `run.sh` provides an abstraction to the tools inside.  Here are an examples of how `run.sh` works as the `ENTRYPOINT`:

````
Self documenting...
$ docker run buildhub:master help


On the CLI or as cronjob: 
$ docker run buildhub:master scrape_archives --auth user:password


As a daemon: 
$ docker run --envfile /etc/dockerflow/buildhub-pulse.txt \
    buildhub:master pulse_listener
````

Web UI
=====
The WebUI (SearchKit: #127) also needs to be deployed. 

- Hosted in S3 fronted by Cloudfront. On the https://buildhub.services.mozilla.com domain? 
- Automatic deployment from Circle => Stage|Prod would be nice to have

mostlygeek:
See this deployment bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1380695

leplatrem:
Since the server side is already shipped as a Docker container, the jobs in a Lambda archive, and the UI deployed on Github (modulo #234), should we consider this done?

leplatrem:
Well, since the ``latest-inventory-to-kinto` takes more than 5min, we can't run it from a Lambda.
I'll give a shot to this.

PR #76, opened 2017-05-31
Travis config

Natim:
The goal of this PR is to add the Travis configuration.
However I did a bit of cleanup and add @glasserc review to rename ``format`` into ``formatFilesize``

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,15 @@ matrix:
     - python: 3.6
       env:
         - TOX_ENV=py36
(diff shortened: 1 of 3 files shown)


mostlygeek:
All the cool kids are using CircleCI now. ;) I would recommend building in circle and shipping a docker container (see #41) so its easy for ops to deploy.

Natim:
We will definitely use CircleCI to build a Docker container when we need it.

Issue #90, opened 2017-06-02
Filtering on the build id with a LIKE only accepts numbers?


magopian:
So it seems that filtering on a build id is also impacted by https://github.com/Kinto/kinto/issues/1217 :

I added quotes around the build id, but then it seems only searches that are numbers give results.
Searching for ["2017"](https://kinto-ota.dev.mozaws.net/v1/buckets/build-hub/collections/releases/records?_limit=10&_sort=-build.date&like_build.id=%222017%22) gives a few results, including `2017-06-01T20:32:00`.

But searching for ["2017-06-01T20:32:00"](https://kinto-ota.dev.mozaws.net/v1/buckets/build-hub/collections/releases/records?_limit=10&_sort=-build.date&like_build.id=%222017-06-01T20%3A32%3A00%22) gives no results. Same for ["2017-"](https://kinto-ota.dev.mozaws.net/v1/buckets/build-hub/collections/releases/records?_limit=10&_sort=-build.date&like_build.id=%222017-%22) or even [just "T"](https://kinto-ota.dev.mozaws.net/v1/buckets/build-hub/collections/releases/records?_limit=10&_sort=-build.date&like_build.id=%22T%22).

@Natim any clue what's happening?

magopian:
Well, that's crap, there's no build id that look like a formatted date/timestamp. My bad.

Issue #89, opened 2017-06-02
Debounce the build id search box


magopian:
At the moment, a request is sent to Kinto on each keypress. We should rather debounce it (only send a request whenever no keypress was detected for 200ms or more, or add a "search" button).

Natim:
I would rather use a search button for this.

n1k0:
+1,

Issue #65, opened 2017-05-30
Remove the *Filter fields from the model


magopian:
Once #40 is landed, we'll have a `Route` stored in the model, which holds all the information for the filters, thus making the individual `*Filter` fields redundant.

n1k0:
> we'll have a Route stored in the model, which holds all the information for the filters, thus making the individual *Filter fields redundant.

But we still have to store selected filters somewhere before sending the form, right? I can't really see us storing a *temporary route* every time users tweak a filter, then actually load that route when they press the submit button. Sure, that might be technically feasible, but slightly overkill I think.

But maybe you has something else in mind?

magopian:
This issue was relevant when changing a filter changed the route (and the results) dynamically. I don't think it's relevant anymore now that we have a form. Closing, will reopen in the future if we get back to something more dynamic.

PR #91, opened 2017-06-02
Fix the build id search... kinda

### LINKED ISSUE #87: Using the build id search box breaks filtering
Whenever there's a string entered in the build id search box, two things happen:
1/ you can't filter on any other filter (selecting any other filter should reset the build id search box)
2/ it doesn't filter on the build id at all

magopian:
Fixes #87

screenshot from 2017-06-02 16-03-29

See #90 for why this isn't totally fixing the build id search.


ui/Main.elm
```diff
@@ -59,35 +59,41 @@ update msg ({ filterValues } as model) =
...
         UpdateFilter newFilter ->
             let
+                noFilters =
+                    { model
+                        | productFilter = "all"
+                        , versionFilter = "all"
+                        , platformFilter = "all"
+                        , channelFilter = "all"
+                        , localeFilter = "all"
+                        , buildIdFilter = ""
```
n1k0: Side remark: it really feels like we should have a `Filters` type, and some `init` function for it. We could then have `FilterMsg`, an update function for it, and a dedicated module. 

Could be done in a next PR though.

magopian:
Yeah we definitely need to debounce it, I submitted #89 for just that.

ui/Main.elm
magopian: Indeed! I had submitted #65 some time ago, but didn't get to it yet, as it seems less urgent

PR #93, opened 2017-06-02
Fix slow loading times caused by a elm-compiler bug https://github.com/elm-lang/elm-compiler/issues/1521

### LINKED ISSUE #92: Very slow loading time, or error messages in the console
I was experiencing very slow loading times from the UI, but thought it was because of my old computer but @Natim and @n1k0 were having the same issues, and sometimes even worse with error messages in the console like http://i.imgur.com/OO36Ymf.png or http://i.imgur.com/cdbaxRF.png

After some investigation, it seems we were hitting https://github.com/elm-lang/elm-compiler/issues/1521

The workaround is simple: split long lists and concatenate them.

magopian:
Fixes #92


n1k0:
Tested the patch locally, solves the issue! Great job.

PR #95, opened 2017-06-05
Fix #94: Move Elm code into src.

### LINKED ISSUE #94: Move Elm code into src
So that searching for Elm bits is easier.

n1k0:


ui/package.json
```diff
@@ -3,16 +3,16 @@
   "version": "0.0.1",
   "description": "Firefox System Build info manager.",
   "scripts": {
-    "build": "elm-make Main.elm --output=build/app.js && npm run copy-assets",
-    "copy-assets": "copyfiles index.html style.css build/",
-    "debug": "elm-live Main.elm --output=app.js --debug",
+    "build": "npm run copy-assets && elm-make src/Main.elm --output=build/app.js",
+    "copy-assets": "rimraf build && mkdir build && cp -r public/* build",
```
magopian: TIL about `rimraf`. Do you want to add it to the npm dependencies so we're sure it's available on all platforms? (Eg osx, windows...?)

ui/package.json
```diff
@@ -3,16 +3,16 @@
...
+    "debug": "elm-live src/Main.elm --warn --dir=public/ --output=public/app.js --debug",
```
magopian: :+1: for adding `--warn`

ui/package.json
n1k0: That's an excellent idea :)

ui/public/app.js
```diff
@@ -0,0 +1,18420 @@
+
```
Natim: I don't think we want to commit this file but rather add it in the .gitignore

n1k0: Damn, you're right. Thanks for noticing!

PR #99, opened 2017-06-05
Display the error messages on the UI

### LINKED ISSUE #79: Display error messages on the UI
At the moment, the error message is just a log in the console.

magopian:
Fixes #79 

screenshot from 2017-06-05 11-31-37

### DIFF
--- a/ui/src/Main.elm
+++ b/ui/src/Main.elm
@@ -26,15 +26,12 @@ update msg ({ filterValues } as model) =
             { model
                 | buildsPager = buildsPager
                 , loading = False
+                , error = Nothing
             }
                 ! []
 
         BuildRecordsFetched (Err err) ->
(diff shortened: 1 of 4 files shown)


magopian:
@n1k0 I don't think we need to have a list of errors, at least for now, as there's only one case where we have errors: when we send a request to kinto. We might need one in the future though, true.

I'll change the PR to still keep rendering the rest of UI.

@Natim there's no need for the Debug.log anymore as it's in the UI. I can add a dismiss button if you think it makes sense, I'll do that in another PR (filed #100)

magopian:
@n1k0 i changed it:

screenshot from 2017-06-05 11-58-36

The server seems to be down at the moment, so the only thing we're rendering is the "no results" ui ;)

Issue #48, opened 2017-05-19
New architecture / design for buildhub?


mostlygeek:
In the Wednesday meeting @magopian mentioned a new structure/design for the WebUI. Can we document the design / decisions for it on GH?

magopian:
Now that #40 has landed, we have the structure in place to link to any build and any "filtered page". Now we need to address the "how do we deal with the massive amount of data", and for that we have #63 to implement pagination.

This in itself should be sufficient to navigate through all the data.

However, what we discussed was that it might be interesting and more ergonomic to change the way the UI displays the list of filters (a big list on the right), to be more like a "catalog" with categories.

Instead of displaying a big list of the latest 100 builds, and then dynamically filter through it, we could instead present the user with a few filters to select:
- product
- channel
- platform
- version
- locale

The user could then choose one (or many) of those, and start filtering from there. A first request would then be sent to Kinto, to retrieve the (paginated) list of builds that match.

This has a few pros and cons:

pros
====

- No useless list/request at the very beginning, no displaying of useless information on start
- less requests to kinto

cons
====

- the list of filters values will have to be "hardcoded", instead of being computed from the current list (narrowing the list down was updating the filter values, keeping only the relevant ones). It'll thus be possible (and very easy) to select filter values that lead to an empty list
- work will be needed to see how to update this list of filter values (if we have a new locale for example)

Those cons may be partially addressable by using elasticsearch to do the faceting, and using that. We still have to think about the "it's easy to use a combination of filters that gives no result, without any help in this regard": do we have a solution? Do we care?

n1k0:
> Now we need to address the "how do we deal with the massive amount of data"

At our standup today, we mentioned that we don't really expect the database to exceed 80k records. **If true**, this kinda makes overkill using smthg like ES. If **not true**, then you can stop reading the content of this comment. 

We discussed several alternative strategies, like doing everything in memory again (as we did when we started the project), or writing a dedicated kinto plugin to help clients retrieving data more easily. 

The only pain point with the kinto plugin approach is if we want real, actual faceted search, aka updating the list of filters computed against the new results, so you can incrementally refine your search against remaining criterias available on the server.

If we do really want faceted search, only ES and the in-memory approaches are likely to work without much hassle. If our tight schedule (end of June) comes into play, in-memory largely wins I believe (but I may be wrong here, the thing is we won't know for sure until we actively work on this, which I'm planning to do tomorrow anyway).

[...]

PR #101, opened 2017-06-05
Fix #89: Update filters so they act like a form.

### LINKED ISSUE #89: Debounce the build id search box

n1k0:
Fixes #89. This patch wraps current filters into a form you submit as a whole everytime you want to refine your current search.

### Pros

- It saves a bunch of http requests
- It reallows filtering build ids along other filters

### Cons 

- Usage is a little less *dynamic* as you need to submit the form to run a search.

Thoughts?

### DIFF
--- a/ui/src/Main.elm
+++ b/ui/src/Main.elm
@@ -94,12 +94,15 @@ update msg ({ filterValues } as model) =
 
                         NewBuildIdSearch value ->
                             { noFilters | buildIdFilter = value }
-
-                updatedModelWithRoute =
-                    { model | route = routeFromFilters updatedModelWithFilters }
             in
-                { updatedModelWithRoute | loading = True }
-                    ! [ newUrl <| urlFromRoute updatedModelWithRoute.route ]
+                updatedModelWithFilters ! []
+
+        SubmitFilters ->
+            let
+                route =
+                    routeFromFilters model
+            in
+                { model | route = route } ! [ newUrl <| urlFromRoute route ]
 
         UrlChange location ->
             let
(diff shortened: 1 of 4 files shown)


n1k0:
Patch has been merged with latest master (error handling) and deployed to production for easier QA.

n1k0:
Ok I'm just realizing that this approach isn't gonna work when we move to facetted search, where we'll have to update filter lists according to results everytime we refine them.

I'm not closing the PR for now as this might serve as a temporary demo.

magopian:
We can have the search on build id "among" the other filters without having to transform the filters as a form. I just didn't think searching for a build id at the same time as other filters was a valid use case, if it is, i'm fine with it.

Regarding transforming the filters as a form, i believe this discussion should be linked with the issue #48. The idea there was to have some "pre filters" that we would force the user to choose before the first list is being displayed.

I'l still not sure how much sense that makes:
1/ not sure it's easier/more ergonomic
2/ not sure it's needed (we don't have that many records, less than 40k, and the current application seems very snappy, at least to me)

n1k0:
I think the key idea behind faceted search is discovery. You start picking a first filter, then another one, and the result list updates accordingly along a refined filter list. That's super convenient when browsing Amazon, but Firefox builds? Honestly, I dunno.

magopian:
Let's land this, and we'll see about faceted search later.

PR #104, opened 2017-06-05
Fix #103: Load filters from dedicated Kinto collections.

### LINKED ISSUE #103: Add a Kinto plugin that would keep an up-to-date collections with the list of products, versions and locales
Compute 3 collections from the releases collection that would provide a way to grab the filters using the elm-kinto client.

That would be another pick at using ES for the demo.

n1k0:
Fixes #103.


n1k0:
Deployed to production for easier QA.

ui/src/Main.elm
```diff
@@ -51,6 +51,33 @@ update msg ({ filterValues } as model) =
         BuildRecordsNextPageFetched (Err err) ->
             { model | error = Just err, loading = False } ! []
 
+        FiltersReceived filterName (Ok { objects }) ->
+            let
+                values =
+                    List.map .name objects
```
Natim: 👍

ui/src/Model.elm
```diff
@@ -33,51 +32,67 @@ init location =
...
+              , getFilters "locale"
+              , getBuildRecordList updatedModel
+              ]
+
+
+getFilters : String -> Cmd Msg
+getFilters filterName =
+    client
+        |> Kinto.getList (filterRecordResource filterName)
+        |> Kinto.sortBy [ "name" ]
```
Natim: I would rather sort by ``id`` I am working on a patch to make sure the version are ordered in the right order (not the string one but really the version one.)

n1k0: Ah, good info, thanks. Will update.

ui/src/Model.elm
```diff
@@ -33,53 +32,52 @@ init location =
...
+    {- FIXME: https://github.com/Kinto/kinto/issues/1217: here we surround all qs param values with quotes
+       in case they're treated as numbers by Kinto, which makes it crash.
+    -}
     let
-        request =
-            client
-                |> Kinto.getList recordResource
-                |> Kinto.limit pageSize
-                |> Kinto.sortBy [ "-build.date" ]
+        applyListFilter apply filter request =
```
magopian: `apply` seems like a poor name, how about `predicate`? Also, why `List` in the `applyListFilter` name? Why not just `applyFilter` (or even `addFilter`)?

Not very important, just a nit ;)

ui/src/Model.elm
```diff
@@ -33,53 +32,52 @@ init location =
...
+            else
+                request
 
-        filteredRequest =
+        applyBuildIdFilter request =
             if buildIdFilter /= "" then
-                request
-                    -- Temporary workaround for https://github.com/Kinto/kinto/issues/1217 : surround version with quotes
-                    |> Kinto.withFilter (Kinto.LIKE "build.id" <| ("\"" ++ buildIdFilter ++ "\""))
+                Kinto.withFilter (Kinto.LIKE "build.id" <| ("\"" ++ buildIdFilter ++ "\"")) request
```
magopian: you've implemented `applyListFilter` above to accept a predicate, but you're not using it here with the `Kinto.LIKE` predicate? And below you're only using the `Kinto.Equal` predicate?

ui/src/Model.elm
```diff
@@ -33,53 +32,52 @@ init location =
...
+        client
+            |> Kinto.getList buildRecordResource
+            |> Kinto.limit pageSize
+            |> Kinto.sortBy [ "-build.date" ]
+            |> applyListFilter (Kinto.Equal "source.product") productFilter
+            |> applyListFilter (Kinto.Equal "target.channel") channelFilter
+            |> applyListFilter (Kinto.Equal "target.platform") platformFilter
+            |> applyListFilter (Kinto.Equal "target.version") versionFilter
+            |> applyListFilter (Kinto.Equal "target.locale") localeFilter
+            |> applyBuildIdFilter
```
magopian: Nice rewrite :+1:

ui/src/Model.elm
n1k0: > Also, why List in the `applyListFilter` name?

Because we're using `getList` just above, but I can totally update the name to `applyFilter`. I'll do that in another PR though, as this one has landed already.

ui/src/Model.elm
n1k0: That's because `buildId` has a different value than `"all"`, and I didn't want to add that argument to `applyListFilter`. I can change that if you really want me to :)

magopian: ah, yes, i understand. I guess we could "factor out" something called `kintoFilter`:

```elm
kintoFilter predicate filterValue =
    Kinto.withFilter (predicate filterValue)
```
But not sure how useful it is.

ui/src/Model.elm
magopian: Ah, ok, makes sense now, but then should the one below be renamed `applyBuildIdListFilter`? :P

ui/src/Model.elm
n1k0: Not sure that'd bring a lot of value to the table, but I can totally do that sure.

ui/src/Model.elm
n1k0: I hesitated, then went for the shorter name.

PR #105, opened 2017-06-05
Add a CLI tool to rebuild the list of filters.

Natim:
Refs #104


Natim:
> Is it because Mozilla uses some super specific version naming conventions?

I discover that way of doing in the socorro database, I found it cool.
To be honest I didn't look at a lib because I found it obvious to code.

>  are we sure this is future proof enough so that this script will parse anything thrown at it from what we plan to aggregate?

Yes I think so but I plan on adding in the tests any use case that we discover along the way.

PR #107, opened 2017-06-05
Move filters to their own type and data structure.

n1k0:


ui/src/Main.elm
```diff
@@ -79,49 +100,12 @@ update msg ({ filterValues } as model) =
...
-                            { noBuildId | channelFilter = value }
-
-                        NewLocaleFilter value ->
-                            { noBuildId | localeFilter = value }
-
-                        NewBuildIdSearch value ->
-                            { noFilters | buildIdFilter = value }
-            in
-                updatedModelWithFilters ! []
+            { model | filters = updateFilters newFilter filters } ! []
```
Natim: Don't you have to use model.filters there?

n1k0: no, the `filters` is destructured in the `update` signature.

PR #116, opened 2017-06-06
First version of the release checking script.

Natim:
This script scrape everything and give info about:

- Files that were skipped
- Files that were found
- Files that were missing

I guess we could add a '--fix-missing' options or something similar but for now it is just informative.


leplatrem:
Don't you think we could merge it with the existing `scrape_archives` script? As it there is loads of duplicated code and I believe we could do better.

As far as I understand, the difference is that:
* it does not check for the latest known product version
* it scans everything
* it queries the server to check if a record already exists or not
* it creates the records that don't exist

We could change the scrape archives script to:
* add a `--full` option
* create the records that don't exist (something like `batch.create_record(data=record, if_not_exists=True)`)

What do you think?

jobs/buildhub/check_missing_releases.py
```diff
@@ -0,0 +1,335 @@
...
+
+    while True:
+        record = await queue.get()
+        release_url = record['download']['url']
+        release_id = build_record_id(release_url)
+        if release_id is None:
+            print('Skipping', release_url)
+            continue
+        try:
+            client.get_record(id=release_id)
```
leplatrem: wouldn't it more efficient to obtain the list of ids at first and put it in a `set()` instead of a hit on the API for each entry?

Natim: It was not supposed to find lots of them. I guess the best way is to create a worker to would consume a list of archive that needs to be added to Kinto and other scrpts that would populate that list. So that we can have different kind of way to populate that list (S3 for instance).

The plan of this script was to validate the work done in `scraping_archive.py` so merging them defeat the purpose of this script I guess.

leplatrem:
Superseded by #185

PR #124, opened 2017-06-07
Change build permalinks to be based on product-platform-version-locale

### LINKED ISSUE #108: Add build permalinks
We should add permalinks for builds, so user can easily share information about a single build.

Previous issue content for the records:

>1. go to https://mozilla-services.github.io/buildhub/#/builds/product/firefox/channel/beta/platform/all/version/54.0b13/locale/en-US
>2. click the second entry permalink (https://mozilla-services.github.io/buildhub/#firefox_beta_54-0b13_linux->x86_64_en-us)
>3. fennec builds are listed, they shouldn't as we want a builds

magopian:
Fixes #108

### DIFF
--- a/ui/src/View.elm
+++ b/ui/src/View.elm
@@ -1,12 +1,13 @@
 module View exposing (view)
 
+import Filesize
 import Html exposing (..)
 import Html.Attributes exposing (..)
 import Html.Events exposing (..)
 import Kinto
 import Snippet exposing (snippets)
 import Types exposing (..)
-import Filesize
+import Url exposing (..)
 
 
 view : Model -> Html Msg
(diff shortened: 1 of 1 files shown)


ui/src/View.elm
```diff
@@ -217,7 +218,21 @@ recordView record =
...
-                    [ a [ href <| "./#" ++ record.id ]
+                    [ a
+                        [ let
+                            url =
+                                { product = record.source.product
+                                , version = record.target.version
+                                , platform = record.target.platform
+                                , channel = "all"
+                                , locale = record.target.locale
+                                , buildId = ""
```
n1k0: How about adding the build id if it's available?

magopian:
I added the info, what do you think?

PR #126, opened 2017-06-07
Fix #125: Skip analyzing fields possibly containing hyphens.

### LINKED ISSUE #125: Allow matching exact filter values
In #117 I'm adding faceted search, so I need to filter values and compute aggregates against string fields containing hyphens, eg `linux-i686`. Right now search will strip the hyphen and match against both `linux` and `i686` values, which is not what we want; what we want is to match against a filed strucly matchin the whole `linux-i686` characters sequence.

To allow that, we need to alter our index schema and set the `index` property to `not-analyzed` for these fields:  https://www.elastic.co/guide/en/elasticsearch/guide/current/_finding_exact_values.html

n1k0:
Attempt at addressing #125. I don't know how to test that. Help welcome.

### DIFF
--- a/jobs/initialization.yml
+++ b/jobs/initialization.yml
@@ -9,6 +9,7 @@ build-hub:
           properties:
             id:
               type: "keyword"
+              index: "not_analyzed"
             last_modified:
               type: "long"
             build:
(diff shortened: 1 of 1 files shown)


jobs/initialization.yml
```diff
@@ -18,36 +19,48 @@ build-hub:
...
                   type: "keyword"
+                  index: "not_analyzed"
                 mimetype:
                   type: "keyword"
+                  index: "not_analyzed"
                 size:
                   type: "long"
                 date:
                   type: "date"
+                  index: "not_analyzed"
```
leplatrem: nit: might night be necessary for type date

n1k0:
I quickly searched in the kinto database where hyphens could come into play when searching and updated the index schema accordingly.

n1k0:
Now would be nice to deploy it and probably reindex the records on production. Could someone help with this? :)

PR #129, opened 2017-06-07
Upgrade elm-kinto

Natim:


Natim:
Apparently the current range already covers the patch release.

PR #135, opened 2017-06-12
Trying searchkit

magopian:
Based on the work from @leplatrem, a few experiments to see how far we can go with searchkit.

- [x] add a link to the revision when it's present
- [x] add missing information for the records
- [x] add missing filters
- [x] change filters to allow multiple selection
- [x] only display important informations for each entry, and then "expand" the entry to display everything

For the last point: by default only a few fields are displayed:

screenshot from 2017-06-13 15-32-31

Clicking on any row will "expand" it:

screenshot from 2017-06-13 15-32-51

Yes indeed it is very ugly, I'm open to any suggestions ;)


magopian:
I believe this PR brings the searchkit version to feature parity with the elm version (and even one ups it by having a expand/collapse to display more information).

Thoughts @n1k0 @leplatrem @mostlygeek ?

n1k0:
Added a `deploy` command in https://github.com/mozilla-services/buildhub/pull/135/commits/7b95d979837d6409a1fef144ba27f00c398bc9d4: https://mozilla-services.github.io/buildhub/searchkit/.

mostlygeek:
SearchKit + ElasticSearch is a good out of the box user experience. 

I'm curious what the dev experience is when we start tweaking/tuning it. 

Played with the demo @n1k0 posted:

1. The hyphen problem (#125 #126 #140) still plagues us. Seems to affect Channel, Platform and Locale. **I think this is a high priority** since it shows confusing data in the UI and hides bad data in the db. 
1. I like how we made Products are tabs 👍 
1. Loading 1000 records is too slow (~1200ms for me). I think 100ms to 250ms should be our target for it to feel more instant. 
1. Sort Fx versions from newest -> oldest
1. Click to expand details is nice. Can we put things in an organized table like the Elm UI? 
1. Some documentation / tips on using search. I searched for a revision ID and got nothing
1. Using View More to expand really long lists is tedious. Especially for the Versions. 
1. The sort drop down on the top right is confusing. The options should at least match column names
1. Is there a way to create custom filters easily, like: `Id: 201706*`?

magopian:
> Loading 1000 records is too slow (~1200ms for me). I think 100ms to 250ms should be our target for it to feel more instant.

The number of records that are configured to be displayed per page are 30. Not sure why we're getting 1000 back. Opened #147

> Sort Fx versions from newest -> oldest

Opened #148

> Click to expand details is nice. Can we put things in an organized table like the Elm UI?

Opened #150

> Some documentation / tips on using search. I searched for a revision ID and got nothing

Yeah, for some reason it needs a `*` to match. Opened #151

> Using View More to expand really long lists is tedious. Especially for the Versions. 

I'm not sure how to deal with that. We could have more versions displayed by default, but it would then use more real estate. The other option would be to have a selectbox instead, but it would prevent selecting multiple versions (unless we implement/use a clever react component that would be a select of checkboxes?). Opened #152 

> The sort drop down on the top right is confusing. The options should at least match column names

Opened #153

> Is there a way to create custom filters easily, like: `Id: 201706*`? 

I think there is, to be experimented. Opened #154

magopian:
Should we land this now? @n1k0 ?

n1k0:
Is it deployed to prod? I'm gonna give the full patch a review.

magopian:
it is! Thanks

ui/elastic/src/App.js
```diff
@@ -0,0 +1,260 @@
...
+              <th>Product</th>
+              <th>Version</th>
+              <th>Platform</th>
+              <th>Channel</th>
+              <th>Locale</th>
+              <th>Published on</th>
+            </tr>
+          </thead>
+          <tbody>
+            {hits.map(hit => {
```
n1k0: Nit: you could probably destructure `source`, `target` and friends from `_source` to reduce verbosity:

```js
{hits.map(({_source: {build, download, source, target}}) => {
```

ui/elastic/src/App.js
```diff
@@ -0,0 +1,260 @@
...
+              const revisionUrl = (hit._source.source.revision)
+                ? (<a href={(hit._source.source.repository + '/rev/' + hit._source.source.revision)}>{hit._source.source.revision}</a>)
```
n1k0: Nit: [template strings](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals) are great

ui/elastic/src/App.js
```diff
@@ -0,0 +1,260 @@
...
+              <SideBar>
+                <RefinementListFilter field="target.version"
+                            title="Version"
+                            id="versions"
+                            size={1000}
+                            operator="OR"
+                            orderKey="_term"
+                            orderDirection="desc"
+                            listComponent={ItemCheckboxList}
+                            translations={{"All":"All versions"}}/>
```
n1k0: Nit: This is very uncommon JSX indentation; how about the more common:

```jsx
<RefinementListFilter 
  field="target.version"
  title="Version"
  id="versions"
  size={1000}
  operator="OR"
  orderKey="_term"
  orderDirection="desc"
  listComponent={ItemCheckboxList}
  translations={{"All":"All versions"}}/>
```

ui/elastic/src/App.js
```diff
@@ -0,0 +1,260 @@
...
+                  <ActionBarRow>
+                    <SelectedFilters/>
+                    <ResetFilters/>
+                  </ActionBarRow>
+
+                  <MenuFilter field="source.product"
+                              title="Product"
+                              id="products"
+                              listComponent={Tabs}
+                              translations={{"All":"All products"}}/>
```
n1k0: Nit: missing space after column in object definition. Also, shouldn't we use [prettier](https://github.com/prettier/prettier) to format code and avoid wasting time with CS bikeshedding?

ui/elastic/src/App.js
```diff
@@ -0,0 +1,260 @@
...
+                </ActionBar>
+
+                <Hits hitsPerPage={30} listComponent={HitsTable(this.toggleExpand.bind(this), this.state.expandedEntry)}/>
```
n1k0: Nit: instead of binding, you can declare this method as a class property so the context will be naturally available in the function. This usually helps when facing rendering performance issues and implementing `shouldComponentUpdate`. Here's an example https://github.com/mozilla-services/react-jsonschema-form/blob/master/src/components/Form.js#L89-L102

ui/elastic/src/App.js
```diff
@@ -0,0 +1,260 @@
...
+                <NoHits translations={{
+                  "NoHits.NoResultsFound":"No release found were found for {query}",
+                  "NoHits.DidYouMean":"Search for {suggestion}",
+                  "NoHits.SearchWithoutFilters":"Search for {query} without filters"}} suggestionsField="target.version"/>
```
n1k0: Nit: missing space after columns.

ui/elastic/src/App.js
magopian: I changed that in https://github.com/mozilla-services/buildhub/pull/135/commits/ae1657aaeae69397c812b48f8d79579631283534 (if I understand your advice correctly).

magopian:
I believe i've addressed all your comments, please let me know if there's something I misunderstood @n1k0 !

n1k0:
All Good

PR #136, opened 2017-06-12
Scrape all available build-ids for all releases.

### LINKED ISSUE #110: Build id information should be a required information
Right now most build records in the database don't seem to have any `build` object attached to them, while it should contains the build id and build date, which we declared critical information for this project.

For instance, we currently only seem to have 43 build records exposing that information: https://kinto-ota.dev.mozaws.net/v1/buckets/build-hub/collections/releases/records?_limit=10&_sort=-build.date&source.product=firefox&not_build.id

I think the build objet should be a required information, and/or we shouldn't list builds without that information. A compromise could probably be to have an option in the search UI to allow or disallow listing builds with that information missing.

Natim:
Fixes #110

### DIFF
--- a/jobs/buildhub/scrape_archives.py
+++ b/jobs/buildhub/scrape_archives.py
@@ -187,6 +187,7 @@ async def fetch_listing(session, url):
 
 _nightly_metadata = {}
 
+
 async def fetch_nightly_metadata(session, product, nightly_url):
     """A JSON file containing build info is published along the nightly archive.
     """
--- a/jobs/buildhub/scrape_archives.py
+++ b/jobs/buildhub/scrape_archives.py
@@ -3,6 +3,7 @@ import async_timeout
 import concurrent.futures
 import datetime
 import logging
+import os.path
 import re
 import sys
 from packaging.version import parse as version_parse
@@ -11,7 +12,7 @@ import aiohttp
 import backoff
 from buildhub.utils import (
     FILE_EXTENSIONS, build_record_id, parse_nightly_filename, is_release_metadata,
-    is_release_filename, guess_mimetype, guess_channel
+    is_release_filename, guess_mimetype, guess_channel, localize_nightly_url
 )
 from kinto_http import cli_utils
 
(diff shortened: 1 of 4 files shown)


jobs/buildhub/utils.py
```diff
@@ -11,6 +11,19 @@
     }
 
 
+def localize_nightly_url(nightly_url):
```
Natim: It is ugly, I know, but tested... And yes scrapping is ugly I guess.

jobs/buildhub/scrape_archives.py
```diff
@@ -180,16 +181,24 @@ def archive_url(product, version=None, platform=None, locale=None, nightly=None,
...
-    # XXX: It is only available for en-US though. Should we use the same for every locale?
-    if "en-US" not in nightly_url:
+    if not is_release_filename(product, os.path.basename(nightly_url)):
         return None
 
+    # Make sure the nightly_url is turned into a en-US one.
+    nightly_url = localize_nightly_url(nightly_url)
+
+    if nightly_url in _nightly_metadata:
+        return _nightly_metadata[nightly_url]
```
glasserc: Is it possible to use a `memoize` decorator on an async function instead of this handwritten one?

jobs/buildhub/utils.py
n1k0: I couldn't agree more (and I wrote a famous scraper).

jobs/buildhub/scrape_archives.py
```diff
@@ -180,16 +181,24 @@ def archive_url(product, version=None, platform=None, locale=None, nightly=None,
...
-async def fetch_nightly_metadata(session, nightly_url):
+_nightly_metadata = {}
+
+async def fetch_nightly_metadata(session, product, nightly_url):
     """A JSON file containing build info is published along the nightly archive.
     """
-    # XXX: It is only available for en-US though. Should we use the same for every locale?
-    if "en-US" not in nightly_url:
+    if not is_release_filename(product, os.path.basename(nightly_url)):
         return None
```
leplatrem: Why do we have to do this here? Can you explain in a comment? Why not do this in the calling code?

jobs/buildhub/utils.py
```diff
@@ -11,6 +11,19 @@
...
+    nightly_url = nightly_url.replace('-l10n', '')
+    parts = nightly_url.split('.')
+    locale = 'en-US'
+    index = -3
+    if 'tar' in nightly_url:
+        index = -4
+    if 'mobile' in nightly_url:
+        locale = 'multi'
+    parts[index] = locale
+    return '.'.join(parts)
```
leplatrem: I usually prefer regexps, but it may be a matter of taste...

Also could you add a comment here to explain what does this function? At least with an example and why we have to do this

jobs/buildhub/scrape_archives.py
Natim: It should be possible yes. Can we do it in another PR?

jobs/buildhub/scrape_archives.py
Natim: Yes we can do it in the calling code.

Natim: So I did it there because `fetch_nightly_metadata` was called with invalid URLs, it made sense for me to validate it inside the function itself.

jobs/buildhub/scrape_archives.py
Natim: Refs https://github.com/mozilla-services/buildhub/issues/144

Issue #139, opened 2017-06-12
Allow multiple values in filters


mostlygeek:
I would like to be able to select multiple values for filters. For example, I would like to select: 

Product: Firefox
Channels: Release
Versions: 53.0, 53.0.2, 53.0.3

n1k0:
Very good idea! Also that's gonna be a good flexibility benchmark between Elm and searchkit 😊

magopian:
It's nearly a one-liner in searchkit: https://github.com/mozilla-services/buildhub/pull/135/commits/47daf9fc12889fb2c73392cd5b98f02c68ff16c9

![screenshot from 2017-06-13 10-25-46](https://user-images.githubusercontent.com/167767/27073065-dc7223c8-5022-11e7-81de-29bc28f88ae8.png)

n1k0:
Ongoing implementation is currently happening in #155.

magopian:
Support for this feature is now live in both frontends (elm and searchkit)

PR #155, opened 2017-06-14
Ref #139, WiP: Allow searching from multiple filter values.

n1k0:
WiP refs #139. This patch is currently continuously deployed to production for easier QA.

### TODO

- [x] Search builds using multiple values for a given filter
- [x] Keep all filter values to allow multiple selection
- [x] Fix routing
- [x] Fix query (#163)
- [x] Order builds by download date DESC
- [x] Refine facets using active filters 
- [x] Add a way to reset an individual filter


n1k0:
Note: some urls already work: https://mozilla-services.github.io/buildhub/#/builds/product/firefox|thunderbird/channel/all/platform/all/version/all/locale/all/buildId//page/1

ui/src/ElasticSearch.elm
```diff
@@ -7,75 +7,179 @@ import Json.Encode as Encode
...
+clauseName : ClauseKind -> String
+clauseName kind =
+    case kind of
+        Prefix ->
+            "prefix"
+
+        Match ->
+            "match"
+
+        _ ->
```
glasserc: Why not just use `Term` here? That way you'll get a heads-up if you add a new `ClauseKind`.

ui/src/Update.elm
```diff
@@ -21,29 +21,43 @@ navigateToPage ({ filters } as model) newPage =
...
+    let
+        toggleFilter facets value values =
+            let
+                newValues =
+                    if List.member value values then
+                        List.filter (\v -> v /= value) values
+                    else
+                        value :: values
+            in
+                if List.length newValues == List.length facets then
```
glasserc: This is to clear all of them if they all get checked? Maybe a comment would be helpful?

ui/src/Url.elm
```diff
@@ -36,11 +36,11 @@ routeFromUrl model location =
...
-                        , product = product
-                        , channel = channel
-                        , platform = platform
-                        , version = version
-                        , locale = locale
+                        , product = product |> String.split "|" |> List.filter (\v -> v /= "all")
+                        , channel = channel |> String.split "|" |> List.filter (\v -> v /= "all")
+                        , platform = platform |> String.split "|" |> List.filter (\v -> v /= "all")
+                        , version = version |> String.split "|" |> List.filter (\v -> v /= "all")
+                        , locale = locale |> String.split "|" |> List.filter (\v -> v /= "all")
```
glasserc: I don't agree with this. I think it should be something more like `let parseFilter textFilter = if textFilter == "all" then [] else String.split "|" textFilter`. Right now a URL containing `55.0b1|56.0a1|all` is parsed but I think it shouldn't be.

This might be a nit, because I guess either way, we can't have a filter value called `all`...

ui/src/Update.elm
n1k0: Yeah, though I'm not super happy with the UX involved... It's surprising to see all checkboxes unchecked when clickink on the last one to have all checked. I'll still try to find a better UX.

magopian: I've tried it, and indeed unchecking everything when you check the last filter is confusing. I believe there's two options there:
1/ if everything is checked, just leave it as is
2/ check everything by default, and add a select all/none button

Searchkit uses the first approach.

ui/src/View.elm
```diff
@@ -451,6 +427,42 @@ spinner =
...
+            in
+                div [ class "checkbox" ]
+                    [ label []
+                        [ input
+                            [ type_ "checkbox"
+                            , value entry.value
+                            , checked active
+                            , onWithOptions
+                                "click"
+                                { preventDefault = True, stopPropagation = False }
```
magopian: is this needed? I don't remember having to deal with that in the past, I simply used a "onClick Msg"

n1k0: This is to prevent the checkox to be actually clicked as in regular DOM. That's how we prevent the last checkbox to be checked when we uncheck them all when all filters have been selected.

n1k0:
Went with a dynamic clear button which does a pretty good job at keeping things simple and usable methinks.

multifilter

PR #157, opened 2017-06-14
Fix thunderbird release names

Natim:


leplatrem:
Should we still leave `.zip` then ?

PR #161, opened 2017-06-15
Fix #150: better styling

### LINKED ISSUE #150: [searchkit] better styling for the "expanded details"
See point 5 in https://github.com/mozilla-services/buildhub/pull/135#issuecomment-308193225

magopian:
Fixes #150

This is a second attempt to improve the styling. It's really not my strong point, so please let me know what you think @n1k0 

screenshot from 2017-06-15 14-56-21

I tried sticking as closely as possible to the styling in the elm version, but with the colors from searchkit.


n1k0:
Much better, yet I'm finding the build details a little messy:



How about having the different sections as distinct columns instead stacking up these tables?

magopian:
I'm not sure I understand what you mean, are you talking about having a single table, and then each section (target, download, build...) being a column, and then in those columns you would have one info per line?

n1k0:
yes

magopian:
Gave it a try, but it seems a bit crowded:

screenshot from 2017-06-15 16-45-06

How could we improve that @n1k0 ?

n1k0:
25% width columns?

magopian:
Sure we can even up the columns, but it's still a lot of information to cram into a small horizontal space, don't you think?

screenshot from 2017-06-15 17-09-56

And obviously feel free to take over, you're way better at design ;)

n1k0:
I went with this, thoughts?

mostlygeek:
The column based approach is very nice. 

There's quite a bit of duplicated information. My suggestions: 

searchkit-tweaks

1. The `Target` column can be squashed into the title.
    * `fennec 56.0a1 (android-arm | nightly | Locale: it)` 
1. The `Build` column is totally redundant 
1. The Build id and revision can also squashed to the title bar: `20170614100332 (b266a8)` 
    * a truncated hash conveys as much info as a full length one for humans... 
1. The mimetype doesn't tell us anything the file extension in the URL doesn't. 
1. Size can be made more human readable, ie: 36MB vs 36137120

magopian:
Good ideas and suggestions, will implement. One catch though: the published date and the build date aren't exactly the same. Should we keep only one of the two? Or maybe change the title one to be the build date (when present), and keep the published date in the "download" section?

n1k0:
I tend to concur that all these information could probably fit as columns in a regular table row.

magopian:
Another try:

screenshot from 2017-06-16 10-24-22

n1k0:
I still feel like everything could actually fit on a single row, with no need to expand. That depends if we plan to add more info in the future or not though.

magopian:
I'll open a new PR using the original layout/styling, and adding more information (without the "expand" feature).

magopian:
Closing this in favor of #164

PR #164, opened 2017-06-16
Another attempt at improving the styling for #150

### LINKED ISSUE #150: [searchkit] better styling for the "expanded details"

magopian:
Fixes #150

screenshot from 2017-06-16 11-12-00

### DIFF
--- a/ui/elastic/package.json
+++ b/ui/elastic/package.json
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "filesize": "^3.5.10",
     "react": "^15.5.4",
     "react-dom": "^15.5.4",
     "rimraf": "^2.6.1",
--- a/ui/elastic/src/App.js
+++ b/ui/elastic/src/App.js
@@ -43,6 +43,7 @@ const HitsTable = (toggleExpand, expandedEntry) => {
             <tr>
               <th></th>
               <th>Product</th>
+              <th>Version</th>
               <th>platform</th>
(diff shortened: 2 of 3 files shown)


magopian:
Thanks for the feedback:

screenshot from 2017-06-16 11-28-34

Thoughts @n1k0 ?

magopian:
Maybe we'd want to also have its own column for the version...

n1k0:
Starts looking really nice. Classic stuff, but well known by users and immediately useful when seeking for a specific info.

> Maybe we'd want to also have its own column for the version...

If we plan on allowing ordering that column eventually, yes.

magopian:
Should we merge this one and close #161 then?

n1k0:
A few ideas to gain some real estate:

- Could we find a shoter date format?
- We could put the build id in the first or second column and remove the `#` link
- This may sound silly but if we could find the three icons for ffx, fennec and tb, we could avoid displaying the product name :D

n1k0:
> Should we merge this one and close #161 then?

Ah sorry you commented while I was writing my previous comment. Up to you, we can always iterate further in other PRs.

magopian:
Regarding the `#` it's the permalink to get back to the specific row. The link on the build id is the link to the revision in the repository. I think we want to keep the permalink somehow.

n1k0:
Ah right indeed.

magopian:
Going to land this now.

PR #165, opened 2017-06-19
Add devedition product (fixes #162)

### LINKED ISSUE #162: Aggregate DevEdition builds
re: https://github.com/mozilla-services/socorro/pull/3811

I believe DevEdition == Beta now. It used to track the Aurora channel but Project Dawn killed that. We may want to back fill the previous builds and add a "Dev edition" channel to the UI.

leplatrem:
I really wish I could have done something cleaner. 

I first tried to manage devedition as a «channel», but the AUS expects `aurora`, and lots of code had to be adapted for this specific case.
I thus followed socorro and consider it a different product (which it is I guess), but since the installer starts with `Firefox` and not `Devedition`, I had to tweak our utils functions


leplatrem:
When running this locally I obtained 2830 records

PR #166, opened 2017-06-20
Make sure only one coroutine initializes the candidates list

leplatrem:
There were race conditions where coroutines read the `_candidates` variable while others are still initializing it. Using a mutex fixes the issue.


jobs/buildhub/scrape_archives.py
```diff
@@ -224,20 +225,27 @@ def archive_url(product, version=None, platform=None, locale=None, nightly=None,
...
-            candidate_version = f.replace("-candidates/", "")
-            builds_url = archive_url(product, candidate_version, candidate="/")
-            build_folders, _ = await fetch_listing(session, builds_url)
-            latest_build_folder = sorted(build_folders)[-1]
-            _candidates[product][candidate_version] = latest_build_folder
+    # Make sure there is only one coroutine populating the _candidates list at a
+    # time, using a mutex.
+    await _lock
+
+    try:
```
glasserc: Can we use `with (await _lock)` instead of the ugly finally?

jobs/buildhub/scrape_archives.py
```diff
@@ -213,6 +213,7 @@ def archive_url(product, version=None, platform=None, locale=None, nightly=None,
 
 
 _candidates = {}
+_lock = asyncio.Lock()
```
glasserc: Actually, come to think of it, maybe we could rename this `candidates_lock`?

PR #170, opened 2017-06-20
Live search box

n1k0:
Feedback wanted, deployed to https://mozilla-services.github.io/buildhub/

live2

PR #171, opened 2017-06-21
Search on all fields

magopian:
Refs #170 : live search on all fields in searchkit

Deployed at https://mozilla-services.github.io/buildhub/searchkit

The difference here with #170 is that there isn't automatic `*` added to the search terms, you need to put them yourself. Searching for `fire` won't return any result, but `fire*` will.

### DIFF
--- a/ui/elastic/src/App.js
+++ b/ui/elastic/src/App.js
@@ -57,24 +57,25 @@ const HitsTable = (toggleExpand, expandedEntry) => {
           </thead>
           <tbody>
             {hits.map(
-              ({ _source: { build, download, source, target }, _id }) => {
+              ({ _source: { build, download, source, target }, _id, highlight }) => {
                 const revisionUrl = source.revision
                   ? <a href={`${source.repository}/rev/${source.revision}`}>
                       {source.revision.substring(0, 6)}
                     </a>
                   : "";
+                const getHighlight = (title, value) => {return {__html: ((highlight && highlight[title]) || value)}}
                 return (
                   <tr key={_id} id={_id}>
                     <td><a href={`#${_id}`}>#</a></td>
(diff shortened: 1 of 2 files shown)


magopian:
as explained above, there's no `*` added automatically (yet? not sure how feasible it is), so you'd need to use `firefox linux*`.

n1k0:
In Elm I did this

```elm
prepareSearchQuery : String -> String
prepareSearchQuery search =
    search
        |> String.split " "
        |> List.map (\w -> "*" ++ w ++ "*")
        |> String.join " "
```

But maybe that's too ugly and we should indeed simply rely on default ES simple search query syntax.

magopian:
Yup, I saw that while reviewing your PR. In searchkit's case, I still haven't found the way to modify the query before sending it. There must be a way, just none that I could find (yet).

magopian:
Added highlighted fields:

screen shot 2017-06-22 at 12 33 00

magopian:
This is now doing fulltext search by default like the elm version \o/ No more need for manually adding '*'.

glasserc:
Does automatically surrounding queries with stars preclude the use of more sophisticated ElasticSearch queries by the user? If so, do we care?

magopian:
depends on what you're meaning by "more sophisticated"? If you're talking about https://mozilla-services.github.io/buildhub/searchkit/?q=source.product%3Afirefo%20build.id%3A2017 for example, it still works (not quite sure why though). And to be honest, I have no idea if we care or not.

I mean, unless we document the elasticsearch query syntax (or put a link there or something), we can't expect people to use it. Do we want to? No idea :/

magopian:
I'm landing this as is, we'll experiment and modify/improve in the future if needed.

PR #175, opened 2017-06-23
Add a link to the Elasticsearch query string syntax documentation

### LINKED ISSUE #154: [searchkit] Is there a way to create custom filters like: `Id: 201706*`?

magopian:
Fixes #154

### DIFF
--- a/ui/elastic/src/App.js
+++ b/ui/elastic/src/App.js
@@ -142,7 +142,7 @@ const fullText = (query, options) => {
   if (!query) {
     return;
   }
-  const fulltextQuery = query
+  const fulltextQuery = query.startsWith("'") ? query.slice(1) : query
     .split(" ")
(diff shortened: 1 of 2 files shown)


magopian:
Ok, I tried something: it's only using the elasticsearch query string syntax (sending the query untouched/unparsed) if it starts with a single quote.

What do you think @glasserc ?

glasserc:
Hmm, I feel like a query syntax like `'foo` is kind of strange and asymmetric looking, but at the same time I don't want to try to overcomplicate the frontend logic to try to guess whether to pass the query through or not, and I'm nervous about a more explicit `elasticsearch:QUERY` syntax because it looks too much like regular ES syntax. I guess I'm +0 on this. The code looks OK too, though it looks like it includes an automatic reformatting pass or something.

PR #176, opened 2017-06-29
Fix #106: Use a regular table for listing records.

### LINKED ISSUE #106: Record list entries are too large
When searching for builds, I'm likely to want a quick glance of what my search just retrieved, mostly product, version, build date and id. We could then have a button to expand entries to inspect a specific build details, or even open a modal window with all the information.

If having all the information visible at once is an issue, we could add a supplementary control to expand and collapse all build entries in the list (but then the modal window route would be discarded).

Thoughts?

n1k0:

### DIFF
--- a/ui/public/style.css
+++ b/ui/public/style.css
@@ -39,6 +39,7 @@ td {
   overflow: hidden;
   white-space: nowrap;
   text-overflow: ellipsis;
+  max-width: 65px;
 }
 
 /* Search form */
--- a/ui/public/style.css
+++ b/ui/public/style.css
@@ -33,6 +33,14 @@ body {
     100% { transform: rotate(360deg); }
 }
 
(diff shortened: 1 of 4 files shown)


ui/public/style.css
```diff
@@ -33,6 +33,15 @@ body {
...
 
+/* Records list table */
+
+td {
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  max-width: 65px;
+}
+
```
magopian: Do you also want to add a max-width to the table itself? It seems from your screenshot that it's displaying below the filter selection?

n1k0: That's actually what this commit fixes :) Limiting the table width has no effect.

Natim:
Nice :)

PR #181, opened 2017-07-04
Fix #180: Collapsible build details.

### LINKED ISSUE #180: ability to view a specific build in ui
The ui at https://mozilla-services.github.io/buildhub/#/builds is great for looking for groups of builds, but there doesn't seem to be a "details" view for the rows in that table.

Is there any way to view the details for a specific row?

Amongst other things, this will help in writing up issues for data problems. Then I can list links to details for builds that illustrate the problem.

n1k0:
Refs #180, patch deployed to gh-pages, cc @willkg

collapse

### DIFF
--- a/ui/src/Update.elm
+++ b/ui/src/Update.elm
@@ -135,7 +135,16 @@ update msg ({ filters, settings } as model) =
             model ! []
 
         ToggleBuildDetails id ->
-            if List.member id model.expanded then
-                { model | expanded = List.filter (\i -> i /= id) model.expanded } ! []
-            else
-                { model | expanded = id :: model.expanded } ! []
+            { model
+                | expanded =
+                    case model.expanded of
+                        Just current ->
+                            if id == current then
+                                Nothing
+                            else
+                                model.expanded
+
+                        Nothing ->
+                            Just id
+            }
+                ! []
--- a/ui/src/Update.elm
+++ b/ui/src/Update.elm
@@ -133,3 +133,9 @@ update msg ({ filters, settings } as model) =
 
         SubmitSearch ->
             model ! []
+
+        ToggleBuildDetails id ->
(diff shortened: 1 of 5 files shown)


n1k0:
@willkg is it okay like this? basically you can share permalinks where receiver can click the eye button to expand the build details. We could also create a new dedicated route to have a single given build displayed with the details expanded by default, though that requires a bit more work (though not that much).

ui/src/View.elm
```diff
@@ -286,15 +291,234 @@ buildUrl { build, source, target } =
...
+                                            "close"
+                                        else
+                                            "open"
+                                       )
+                            ]
+                            []
+                        ]
+                    ]
+                , td [ title source.product ]
+                    [ highlighSearchTerm filters.search filters.product source.product ]
```
magopian: Just realised that there's a typo in the name, it's missing a "t": `highlightSearchTerm`

n1k0:
I was also wondering: which info is present in the details view, that isn't present in the summary view?

We're missing system addon versions, that's why we need to expand a record with these details. The thing is, we don't have this information available in the database *yet*. @leplatrem suggested on IRC that we'll eventually have system addons in a separate collection. The good thing with an expandable view is that we will be able to fetch this information when an entry is expanded.

ui/src/Update.elm
```diff
@@ -135,7 +135,16 @@ update msg ({ filters, settings } as model) =
...
+                | expanded =
+                    case model.expanded of
+                        Just current ->
+                            if id == current then
+                                Nothing
+                            else
+                                model.expanded
+
+                        Nothing ->
+                            Just id
```
n1k0: Note: I went with this because swapping expanded entry was making the UI confusing and jiggling.

PR #185, opened 2017-07-06
S3 inventory

leplatrem:
POC here.
* a generic script that reads records as JSON from stdin and publish them on server concurrently
* a specific s3 inventory script that reads the csv, fetch build ids from archive.m.o and outputs records as JSON to stdout

```
$ load-inventory filename.csv | to-kinto --server=https://... --bucket=bid --collection=cid --auth=user:pass
```


jobs/buildhub/to_kinto.py
```diff
@@ -0,0 +1,173 @@
...
+    with client.batch() as batch:
+        for record in records:
+            if "id" in record["data"]:
+                batch.update_record(**record)
+            else:
+                batch.create_record(**record)
+    results = batch.results()
+
+    # Batch don't fail with 4XX errors. Make sure we output a comprehensive
+    # error here when we encounter them.
```
leplatrem: I have having to do this here. The kinto client should offer an option `with client.batch(silent=False)`

leplatrem:
I took code from #116 and from scrape_archives and put it all in one place. I find the result quite readable.
There is still a lot work to do like:
* [ ] import tests from #116 
* [ ] get rid of scrape_archives and mentions
* [ ] get rid of `parse_nightly_filename()` util
* [ ] Ideally, unit test metadata fetching

jobs/buildhub/inventory_to_records.py
```diff
@@ -0,0 +1,213 @@
...
+
+@backoff.on_exception(backoff.expo,
+                      asyncio.TimeoutError,
+                      max_tries=NB_RETRY_REQUEST)
+async def fetch_json(session, url):
+    headers = {
+        "Accept": "application/json",
+        "User-Agent": "BuildHub;storage-team@mozilla.com"
+    }
+    with async_timeout.timeout(TIMEOUT_SECONDS):
```
Natim: Can we catch the TimeOut exception there and print the URL that failed on stderr?

jobs/buildhub/inventory_to_records.py
```diff
@@ -0,0 +1,213 @@
...
+        logger.debug("GET {}".format(url))
+        async with session.get(url, headers=headers, timeout=None) as response:
+            return await response.json()
+
+
+async def fetch_listing(session, url):
```
Natim: Can we avoid doing that? Since we have access to the CSV, can we use it to find out the filename?

Natim:
I would probably add a test for the Timeout too.

jobs/buildhub/inventory_to_records.py
leplatrem: I think that would require to check the whole strategy of this script to load the CSV into memory, but yes we could...

leplatrem: I changed a bit the behaviour. It now scans the candidates once and explicitly on startup (takes ~2sec).

leplatrem:
> I would probably add a test for the Timeout too.

Done :)

leplatrem:
@Natim I won't be able to work on it today. Feel free to take over if you like, it seems we now have the S3 files :) Otherwise tomorrow ;)

jobs/buildhub/inventory_to_records.py
```diff
@@ -0,0 +1,244 @@
...
+                await process_batch(session, batch, stdout)
+
+                batch = []  # Go on.
+
+        await process_batch(session, batch, stdout)  # Last loop iteration.
+
+
+async def main(loop):
+    parser = argparse.ArgumentParser(description='Load S3 inventory as Kinto releases.')
+    parser.add_argument('filename', help='CSV file to load')
```
Natim: It could be nice to be able to handle STDIN here so that we can pipe:

```zcat inventory.csv.gz | inventory_to_record | to_kinto...```

File: jobs/buildhub/utils.py [hunk 5]
```diff
@@ -185,8 +206,9 @@ def record_from_url(url):
     channel = guess_channel(url, version)
 
     if '-' in locale:
-        lang, region = locale.split('-')
-        locale = '{}-{}'.format(lang, region.upper())
+        locale_parts = locale.split('-')
+        locale_parts[1] = locale_parts[1].upper()
+        locale = '-'.join(locale_parts)  # fr-FR, ja-JP-mac
```
leplatrem: Can we add a test for this please? :)

jobs/buildhub/utils.py
```diff
@@ -109,10 +109,24 @@ def is_release_filename(product, filename):
         product = "firefox"
     match_filename = filename.replace(' ', '-').lower()
     re_filename = re.compile("{}-(.+)({})$".format(product, FILE_EXTENSIONS))
-    re_exclude = re.compile(".+(sdk|tests|crashreporter|stub)")
+    re_exclude = re.compile(".+(sdk|tests|crashreporter|stub|gtk2.+xft|-source)")
     return re_filename.match(match_filename) and not re_exclude.match(match_filename)
 
 
+def is_release_url(product, url):
```
leplatrem: Since it is used in the same place as `is_release_filename()` and takes the same argument, I believe we can combine both

jobs/buildhub/inventory_to_records.py
```diff
@@ -145,6 +147,9 @@ def read_csv(stream):
 
     url = archive_url(product, version, platform, locale, candidate=latest_build_folder)
 
+    if 'eme' in url:
+        url = url.replace('eme', 'EME')
```
leplatrem: Why not do `platform = platform.replace(..)` directly in `archive_url()` ?

leplatrem: Also, can you add a test please?

jobs/buildhub/inventory_to_records.py
```diff
@@ -53,6 +53,8 @@ def read_csv(stream):
 
 
 async def fetch_listing(session, url):
+    if 'eme' in url:
+        url = url.replace('eme', 'EME')
```
leplatrem: I'm surprised to see this here. Is it a leftover?

Natim:
https://archive.mozilla.org/pub/firefox/candidates/47.0.1-candidates/build1/linux-i686/en-US/ appear to not exists ? :(

Natim:
Other cases to handle:

- https://archive.mozilla.org/pub/firefox/releases/3.0.16-real/linux-i686/mn/firefox-3.0.16.tar.bz2
- https://archive.mozilla.org/pub/firefox/releases/3.0.19-real-real/linux-i686/th/firefox-3.0.19.tar.bz2
- https://archive.mozilla.org/pub/firefox/releases/24.0-funnelcake24/linux-i686/en-US/firefox-24.0.tar.bz2

Natim:
- https://archive.mozilla.org/pub/firefox/nightly/2010/07/2010-07-04-05-mozilla-central/firefox-4.0b2pre.en-US.win64-x86_64.zip

leplatrem:
The whole thing ran successfully on our dev instance, I'll merge.

PR #187, opened 2017-07-11
Fix esr channel detection for Firefox.

mozbhearsum:
I was looking over the buildhub code and discovered that it doesn't detect "channel" for Firefox ESRs correctly - it sets it to "release" instead of "esr". This patch fixes that, but still needs to be updated to either make it work for Thunderbird (or fix the tests if it already does).

Thunderbird is a bit of a weird case because their "release" channel is built out of ESR repositories - so it's not exactly comparable to Firefox.

### DIFF
--- a/jobs/tests/test_utils.py
+++ b/jobs/tests/test_utils.py
@@ -98,7 +98,7 @@ RECORDS = [
 
     # Firefox ESR
     {
-        "id": "firefox_52-0esr_linux-x86_64_en-us",
+        "id": "firefox_esr_52-0esr_linux-x86_64_en-us",
         "source": {
             "product": "firefox",
         },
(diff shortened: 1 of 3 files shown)


Natim:
That's something we went back and forth about but the current code should detect esr: https://github.com/mozilla-services/buildhub/blob/3ec620c7763cacb95511c0b1199d797a2cef485a/jobs/buildhub/utils.py#L23

Natim:
Thanks

jobs/tests/test_utils.py
```diff
@@ -170,15 +170,15 @@
...
+        "id": "thunderbird_esr_10-0-12esr_macosx_pt-br",
         "source": {
             "product": "thunderbird",
         },
         "target": {
             "version": "10.0.12esr",
             "platform": "macosx",
             "locale": "pt-BR",
-            "channel": "release"
+            "channel": "esr"
```
mozbhearsum: Sorry if I wasn't clear before -- Thunderbird builds still use "release" as their channel, despite being built out of the ESR repositories.

mozbhearsum: I'm pretty sure that .json file is a lie...we use "release" in Balrog. I'm going to double check an actual build to be sure.

Natim: Oh you are right, that's why we went back and forth about it thanks, I will rework that.

PR #191, opened 2017-07-31
Sort versions on major/minor/alpha/beta/pre...

### LINKED ISSUE #149: Find a way to order versions numerically and not alphabetically
There's various ways to do that:
1/ store a "orderable version" (see https://github.com/mozilla-services/buildhub/issues/133#issuecomment-307763965)
2/ use the proper collation (see https://www.elastic.co/guide/en/elasticsearch/guide/current/sorting-collations.html)
3/ sort using a script, aka computed field sorting (see http://www.elasticsearch.org/guide/reference/api/search/sort.html : sorting with a script)

magopian:
Fixes #149

Check the comments in #149 to see how it's done elsewhere

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -1,0 +1,11287 @@
+{
+  "name": "elastic",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@types/cheerio": {
+      "version": "0.22.2",
+      "resolved": "https://registry.npmjs.org/@types/cheerio/-/cheerio-0.22.2.tgz",
+      "integrity": "sha512-7YJQbllem83umRzZ+ry6RZFAbBWNoHvbiW6tE5VEAJIoX6F22hrKrBw3ibRvHm2zPnOhZGrZJusAbp1zZKcLzQ=="
+    },
+    "@types/classnames": {
+      "version": "0.0.32",
+      "resolved": "https://registry.npmjs.org/@types/classnames/-/classnames-0.0.32.tgz",
+      "integrity": "sha1-RJq82agmgHgR7xAeWN+fg8/GFxM="
+    },
+    "@types/enzyme": {
+      "version": "2.8.6",
+      "resolved": "https://registry.npmjs.org/@types/enzyme/-/enzyme-2.8.6.tgz",
+      "integrity": "sha512-nE5lOWlmTaE+of9Ze3en7T/UdlJvhcBVHAH7wc7E0Er09RG/R9g8nAfTpQnJ4hrBvlHxwBkHfe8bewMCD3idDw==",
+      "requires": {
+        "@types/cheerio": "0.22.2",
+        "@types/react": "0.0.0"
+      }
+    },
+    "@types/es6-promise": {
+      "version": "0.0.32",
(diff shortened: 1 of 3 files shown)


ui/elastic/src/App.js
```diff
@@ -89,9 +89,33 @@ const HitsTable = ({ hits }) => {
...
-    const majorB = parseInt(b.key.split(".", 1)[0], 10)
-    return majorB - majorA
+    const partsA = a.key.split(".")
+    const partsB = b.key.split(".")
+    if (partsA[0] !== partsB[0]) {
+      // sort on major version.
+      return parseInt(partsB[0], 10) - parseInt(partsA[0], 10)
+    }
+    if (partsA[1] !== partsB[1]) {
+      const subPartRegex = /^(\d+)([a-zA-Z]+)?(\d+)?([a-zA-Z]+)?/ // Eg: 0b12pre
```
glasserc: It seems like this enforces that `0b12pre` can only happen in after the first period and not after a second, but you also report seeing `38.0.5b1-2`. Is this OK?

ui/elastic/src/App.js
```diff
@@ -89,9 +89,33 @@ const HitsTable = ({ hits }) => {
...
+        return parseInt(subPartB[1], 10) - parseInt(subPartA[1], 10)
+      }
+      if (subPartA[2] !== subPartB[2]) {
+        if (subPartA[2] && !subPartB[2]) {
+          return 1
+        }
+        if (subPartB[2] && !subPartA[2]) {
+          return -1
+        }
+        return subPartA[2] - subPartB[2]
```
glasserc: Subtracting `"a"` from `"b"` produces `NaN` in JS. What is this meant to cover?

magopian: crap.

ui/elastic/src/App.js
magopian: Yeah, i'm trying to balance between "implement all the things so it's perfect" and "make it as simple as possible but still work for very old/legacy versions". For this current case, it seems it just "works well":

screen shot 2017-08-01 at 09 47 18

So I'm not sure I should complicate things a bit more? Another solution would be to use the same "clever parsing for the part after the first period" for each part, and recurse/loop over all the parts. Would that result in cleaner and more understandable code?

glasserc: I think I'd be happy just making it so that it applies to the last element rather than the second element.

magopian:
@glasserc i've changed the implementation, let me know what you think!

ui/elastic/src/App.js
```diff
@@ -132,11 +132,41 @@ const HitsTable = ({ hits }) => {
...
-  });
-};
+    const partsA = a.key.split(".")
+    const partsB = b.key.split(".")
+    if (partsA.length < 2 || partsB.length < 2) {
+      // Bogus version, list is last.
+      return 1;
+    }
+    let i = 0;
+    while (partsA[i] === partsB[i]) { // Skip all the parts that are equal.
```
glasserc: Should you check that you don't run off the end of both? (Otherwise, won't you get an infinite loop if you have two elements that are exactly equal?)

magopian: good catch ;)

PR #201, opened 2017-08-03
Remove system addons from schema

leplatrem:


mostlygeek:
what's the reason for removing system addons from the schema? Are we moving the information somewhere else?

Natim:
> what's the reason for removing system addons from the schema?

Because we currently don't have a way to get that information.

leplatrem:
Currently, we can easily obtain information about available updates by querying Balrog, but the built-ins one is a nightmare. We also realized that it would make more sense to have a distinct collection than having them as attributes of the builds.

I'll open a dedicated issue about it.

PR #209, opened 2017-08-08
Remove Elm implementation of the UI in favor of searchkit

### LINKED ISSUE #208: Tag and remove Elm code
I was told that we're not going to be shipping Elm code. We should create an `elm-ui` tag and delete the code from the repo. If we ever come back to Elm a tag will make it easy to find and resurrect the code.

magopian:
Fixes #208


mostlygeek:
Should I be worried about the failed CI results?

magopian:
The failures have nothing to do with this PR, as they're testing the python code. I'm landing this.

Much sad.

PR #213, opened 2017-08-16
Lambda to load S3 inventories  (fixes #197)

### LINKED ISSUE #197: Add a script/command to load the inventory
Currently it is documented as several commands on the README : https://github.com/mozilla-services/buildhub/tree/master/jobs#s3-inventory

We can do better :) 

A command should:
* list available inventories
* download the manifest of the latest one
* download the .gz files
* run something like `zcat 2017-08-02T00-11Z/*.gz  | grep -v seamonkey tinderbox... | inventory-to-records | to-kinto --skip ...`
* delete the .gz files (?)
* repeat for firefox and archive

**Maybe** it makes sense to do it in python so that the command gets installed with the egg (instead of an isolated `.sh`?)

leplatrem:
fixes #197

* [x] Change existing code to rely on generators
* [x] Implement lambda that chains generators (download, gunzip, parse csv, obtain metadata, load into kinto)
* [x] Refactor `to_kinto.py::main()` in order to take a kinto http client instance instead of raw CLI args
* [x] Add test for the new lambda
* [x] Fix tests to adjust to new spec
* [x] Moved `async def stream_as_generator(loop, stream)` to utils
* [x] Add debug logging

> **Note**: This PR drops support of Python 3.5 because we now use async generators of Python 3.6.

### DIFF
--- a/jobs/buildhub/lambda_s3_inventory.py
+++ b/jobs/buildhub/lambda_s3_inventory.py
@@ -1,5 +1,6 @@
 import asyncio
 import json
+import logging
 import os.path
 import zlib
 
@@ -15,6 +16,9 @@ REGION_NAME = 'us-east-1'
 BUCKET = "net-mozaws-prod-delivery-inventory-us-east-1"
 FOLDER = "public/inventories/net-mozaws-prod-delivery-{inventory}/delivery-{inventory}/"
 CHUNK_SIZE = 1024
+
+
+logger = logging.getLogger()  # root logger.
 
 
 async def list_manifest_entries(loop, client, inventory):
@@ -75,6 +79,10 @@ async def main(loop, event, inventory):
 
 
 def lambda_handler(event, context):
+    # Log everything to stderr.
+    logger.addHandler(logging.StreamHandler())
+    logger.setLevel(logging.DEBUG)
+
     loop = asyncio.get_event_loop()
     for inventory in ("firefox", "archive"):
         loop.run_until_complete(main(loop, event, inventory))
--- a/jobs/buildhub/lambda_s3_inventory.py
+++ b/jobs/buildhub/lambda_s3_inventory.py
@@ -1,0 +1,62 @@
+import asyncio
+import aiobotocore
+import gzip
+import json
+import os.path
+from io import BytesIO, StringIO
+
+from buildhub.inventory_to_records import csv_to_records
+
+BUCKET = "net-mozaws-prod-delivery-inventory-us-east-1"
+FOLDER = "public/inventories/net-mozaws-prod-delivery-{inventory}/delivery-{inventory}/"
+
+
+async def handle_file(loop, client, file_key):
+    # 2. For each inventory file, convert it to records and populate Kinto
+    gzipped_csv_file = await client.get_object(Bucket=BUCKET, Key=file_key)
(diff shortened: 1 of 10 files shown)


jobs/buildhub/lambda_s3_inventory.py
```diff
@@ -0,0 +1,98 @@
...
+    kinto_auth = tuple(os.getenv("AUTH", "user:pass").split(":"))
+
+    kinto_client = kinto_http.Client(server_url=server_url, auth=kinto_auth,
+                                     bucket=bucket, collection=collection,
+                                     retry=NB_RETRY_REQUEST)
+
+    session = aiobotocore.get_session(loop=loop)
+    boto_config = botocore.config.Config(signature_version=botocore.UNSIGNED)
+    async with session.create_client('s3', region_name=REGION_NAME, config=boto_config) as client:
+
```
Natim: spare lines

jobs/buildhub/lambda_s3_inventory.py
```diff
@@ -0,0 +1,98 @@
...
+        keys_stream = list_manifest_entries(loop, client, inventory)
+        csv_stream = download_csv(loop, client, keys_stream)
+        records_stream = csv_to_records(loop, csv_stream)
+
```
Natim: ditto

jobs/buildhub/lambda_s3_inventory.py
```diff
@@ -0,0 +1,98 @@
...
+
+def lambda_handler(event, context):
+    # Log everything to stderr.
+    logger.addHandler(logging.StreamHandler())
+    logger.setLevel(logging.DEBUG)
+
+    loop = asyncio.get_event_loop()
+    for inventory in ("firefox", "archive"):
+        loop.run_until_complete(main(loop, inventory))
```
Natim: Don't you want to use an asyncio.gather here?

jobs/buildhub/lambda_s3_inventory.py
```diff
@@ -0,0 +1,98 @@
...
+    loop.close()
+
+
+if __name__ == "__main__":
+    lambda_handler(None, None)
```
Natim: Can we remove that and make it an entry point in the setup.py instead? (Making None the default value of lambda_handler)

jobs/buildhub/to_kinto.py
```diff
@@ -97,20 +100,10 @@ def publish_records(client, records):
     return results
 
 
-async def produce(loop, queue):
-    """Reads JSON the stdin asynchronuously where each line is a record.
+async def produce(loop, records, queue):
+    """Reads an asynchronuous generator of records and puts them into the queue.
```
Natim: asynchronous

jobs/buildhub/utils.py
```diff
@@ -220,6 +221,30 @@ def chunked(iterable, size):
...
 
+async def split_lines(stream):
+    leftover = ''
+    async for chunk in stream:
+        chunk_str = chunk.decode("utf-8")
+        chunk_str = leftover + chunk_str
+        chunk_str = chunk_str.lstrip("\n")
+        lines = chunk_str.split("\n")
+        leftover = lines.pop()
+        if lines:
```
Natim: Maybe add a comment to say that the end of the line is in the next chunk.

PR #217, opened 2017-08-18
Document S3 inventory lambda

leplatrem:


jobs/README.rst
```diff
@@ -22,8 +22,57 @@ The JSON schema validation can be enabled on the server with the following setti
 
     kinto.experimental_collection_schema_validation = true
 
-S3 inventory
-============
+
+Load latest S3 inventory
+========================
+
+A command downloads the latest S3 manifests from Mozilla archives, and create records on a remote Kinto server.
```
glasserc: `A command to download...`

jobs/README.rst
```diff
@@ -22,8 +22,57 @@ The JSON schema validation can be enabled on the server with the following setti
...
+Its configuration is read from environment variables:
+
+* ``SERVER_URL`` (default: ``http://localhost:8888/v1``)
+* ``BUCKET`` (default: ``build-hub``)
+* ``COLLECTION`` (default: ``releases``)
+* ``AUTH`` (default: ``user:pass``)
+* ``NB_RETRY_REQUEST`` (default: ``3``)
+* ``TIMEOUT_SECONDS`` (default: ``300``)
+
+The following entry point can be used as an Amazon Lambda function:
```
glasserc: Because the name of the lambda is so different, I didn't believe this was the same as the above. How about changing to something like `To use this script as an Amazon Lambda function, use the entry point ``buildhub.lambda_s3_inventory.lambda_handler`` ` ?

PR #218, opened 2017-08-18
Handle devedition builds (fixes #188)

### LINKED ISSUE #188: Handle DevEdition builds correctly
DevEdition is a bit weird. Up to and including 54.0, DevEdition was simply the branding of the Firefox aurora channel builds. Beginning with 55.0, DevEdition became a second build of Firefox that we do from the mozilla-beta repository. This means that it now uses "candidates" and "releases" style directories (see https://archive.mozilla.org/pub/devedition/). For Reasons (tm), it ended up in a different location on archive.mozilla.org, but despite the directory name it should be treated as a "Firefox" build, and is still currently using "aurora" as its update channel.

leplatrem:
Fixes #188 

@mozbhearsum in the issue description you said we should treat devedition builds as firefox builds. But in that case how should distinguish them from firefox aurora builds?

### DIFF
--- a/jobs/buildhub/utils.py
+++ b/jobs/buildhub/utils.py
@@ -4,7 +4,7 @@ import os.path
 import re
 
 
-ALL_PRODUCTS = ("firefox", "thunderbird", "mobile")
+ALL_PRODUCTS = ("firefox", "thunderbird", "mobile", "devedition")
 ARCHIVE_URL = "https://archive.mozilla.org/"
 DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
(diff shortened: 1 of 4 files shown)


mozbhearsum:
I think you can treat them as the same. Before a certain point in time, DevEdition was built out of the mozilla-aurora repository, and uploaded to https://archive.mozilla.org/pub/firefox/nightly/ into directories with "mozilla-aurora" in their names. After a certain point in time, we started building it from mozilla-beta and uploading it to https://archive.mozilla.org/pub/devedition/.  There might be a small amount overlap period where we uploaded to both locations. I realize it might be a pain for DevEdition to be considered both of these things, but that's how it's been viewed from a Product standpoint.

PR #220, opened 2017-08-21
Fix building of manifest key (ref #197)

leplatrem:
ref #197


leplatrem:
The values returned in the `CommonPrefixes` attribute actually contain the absolute key path, and not the relative one from `paginate(Key=prefix)`. 

This patch prevents the final manifest key to contain the prefix twice

PR #228, opened 2017-08-23
Rework documentation (fixes #216)

### LINKED ISSUE #216: Document lambda configuration
basicinfo
lambda
lambdaname
tags-settings

leplatrem:
* [x] Move as much as possible to readthedocs
* [x] Update jobs schema
* [x] Add documentation about Lambda (ref #216)


docs/api.rst
```diff
@@ -0,0 +1,267 @@
...
+Basic API
+---------
+
+The list of records is available at:
+
+* `$SERVER/buckets/build-hub/collections/releases/records <https://kinto-ota.dev.mozaws.net/v1/buckets/build-hub/collections/releases/records?_limit=10>`_
+
+And a single record at:
+
+* `$SERVER/buckets/build-hub/collections/releases/records/${ID} <https://kinto-ota.dev.mozaws.net/v1/buckets/build-hub/collections/releases/records/firefox_beta_50-0b11_macosx_el>`_
```
glasserc: `${SERVER}` to match `${ID}`? Or vice versa?

docs/api.rst
```diff
@@ -0,0 +1,267 @@
...
+-----------------
+
+An ElasticSearch endpoint is also available for more powerful queries. It powers the online catalog.
+
+* `$SERVER/buckets/build-hub/collections/releases/search <https://kinto-ota.dev.mozaws.net/v1/buckets/build-hub/collections/releases/search>`_
+
+`More information in the Elasticsearch documentation <https://www.elastic.co/guide/en/elasticsearch/reference/current/search.html>`_
+
+
+Queries examples
```
glasserc: Maybe something like `Example queries` or `Sample queries`, or even something like `Buildhub Query Cookbook`?

docs/api.rst
```diff
@@ -0,0 +1,267 @@
...
+================
+
+Is this an official build id?
+-----------------------------
+
+In order to check that a build id exists, we'll just check that it is mentionned in at least one record.
```
glasserc: `mentioned`

docs/api.rst
```diff
@@ -0,0 +1,267 @@
...
+.. code-block:: python
+
+    import kinto_http
+
+    client = kinto_http.Client("https://kinto-ota.dev.mozaws.net/v1")
+    records = client.get_records(**{"build.id": "20110110192031", "_limit": 1, "pages": 1},
+                                 bucket="build-hub", collection="releases")
+    print(len(records) == 1)
+
+What is the revision of a build id?
```
glasserc: Does `revision` here mean something like a Mercurial commit ID, or prerelease number, or what?

docs/api.rst
```diff
@@ -0,0 +1,267 @@
...
+    records = client.get_records(**{"build.id": "20110110192031", "_limit": 1, "pages": 1},
+                                 bucket="build-hub", collection="releases")
+    try:
+        revision = records[0]["source"]["revision"]
+    except IndexError:
+        raise ValueError("Unknown build id")
+    except KeyError:
+        raise ValueError("Unknown revision")
+
+What are the locales of a specific version?
```
glasserc: Maybe something like `What locales are available for a certain version?` or `In what locales is a specific version available?`

docs/api.rst
```diff
@@ -0,0 +1,267 @@
...
+
+Here are some example of release ID:
+
+- ``firefox_nightly_2017-05-03-03-02-12_55-0a1_win64_en-us``
+- ``thunderbird_52-0-1_linux-x86_64_en-us``
+- ``firefox_aurora_54-0a2_macosx_en-us``
+- ``firefox_beta_52-0b6_linux-x86_64_en-us``
+- ``firefox_50-0rc1_linux-x86_64_fr``
+- ``firefox_52-0esr_linux-x86_64_en-us``
+- ``fennec_nightly-old-id_2017-05-30-10-01-27_55-0a1_android-api-15_multi``
```
glasserc: I think this whole section probably belongs at the top, along with a little information about what kinds of records are available for a given version (i.e. it seems that there's one record available per version x platform x locale combination, so you'll get several records for a given release).

docs/jobs.rst
```diff
@@ -0,0 +1,147 @@
...
+
+In order to build the AWS Lambda Zip archive in an isolated environment, we use Docker:
+
+* ``make get_zip``
+
+
+Load latest S3 inventory
+========================
+
+A command to download the latest S3 manifests from Mozilla archives, and create records on a remote Kinto server.
```
glasserc: This might be outside the scope of this PR because so far it seems like you're just moving stuff around, but it might be good to add a line or two about what the contents of the S3 manifests are. For example, `A command to download the latest S3 manifests, containing information about all releases available on archive.mozilla.org, and send that information as Buildhub records to a remote Kinto server`. Actually I think the manifest doesn't even contain the information about the releases available, it's just a list of filenames, which we then scrape, right?

docs/jobs.rst
```diff
@@ -0,0 +1,147 @@
...
+
+* ``buildhub.lambda_s3_inventory:lambda_handler``
+
+
+S3 Event lambda
+===============
+
+The Amazon Lambda function that is in charge of keeping the database up-to-date.
+
+When releases are published on S3, a `S3 Event <http://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html>`_ is triggered and `the lambda is invoked <http://docs.aws.amazon.com/lambda/latest/dg/with-s3.html>`_.
```
glasserc: I'd say `an S3 Event`.

docs/api.rst
leplatrem: Didn't get that...

glasserc: Both `$SERVER` and `${ID}` are variable interpolations, but they have different syntax and I don't think they need to, so I'd suggest changing them both to `$FOO` or both to `${FOO}`.

leplatrem: Ohhh I wasn't awake enough! Thank you!

PR #235, opened 2017-08-29
Include server-bid-cid in previous dump file

leplatrem:


jobs/buildhub/to_kinto.py
```diff
@@ -46,6 +46,11 @@ def fetch_existing(client, cache_file=PREVIOUS_DUMP_FILENAME):
     """Fetch all records since last run. A JSON file on disk is used to store
     records from previous run.
     """
+    cache_file = cache_file.format(
+        server="".join(c for c in client.session.server_url if c.isalnum()),
```
Natim: How will that lookslike? `httpbuildhubstagemozawsnetv1` ?

leplatrem: yep

Natim: Maybe the hostname would be more readable?

leplatrem: why would we need it to be more readable?

PR #238, opened 2017-08-30
Use with buildhub stage.

Natim:


leplatrem:
Well, this is problematic if the bucket name is not the same one between servers....

Natim:
> Well, this is problematic if the bucket name is not the same one between servers....

Why ? Can't it be part of the configuration?

leplatrem:
> Why ? Can't it be part of the configuration?

It is a source of confusion, let's keep `buildhub` everywhere

leplatrem:
The stage is slowly being populated in the right bucket. I believe we can merge/deploy this.

PR #245, opened 2017-09-12
Move lambda upload script to jobs

leplatrem:


Natim:
Why would you do that? Then why not move the Dockerfile and the Makefile as well?

leplatrem:
Yeah, you're right, I thought we would put everything *jobs related* in the `jobs/` folder, but nevermind :)

PR #248, opened 2017-09-13
Containerize commands (fixes #41)

### LINKED ISSUE #41: Operational Spec for BuildHub Version 1.0

leplatrem:
* [x] build the lambda from container
* [x] ship a container with all server based code
* [x] follow dockerflow
* [x] Build/test container in circle.ci
* [ ] Publish container to hub.docker.com/mozilla/buildhub
* [x] A single container with all of the scripts / workers / scrapers / etc
* [x] version.json written into the container, accessible by run.sh version
* [x] a run.sh file as the container's ENTRYPOINT


.gitignore
```diff
@@ -7,3 +7,4 @@ jobs/.coverage
 docs/_build/
 .venv/
 lambda.zip
+version.json
```
Natim: missing new line at the end of the file.

leplatrem:
The remaining step consists in enabling the repo in CircleCI.

Also, I didn't setup the unit tests task in CircleCI because I was running into permissions issues (the `app` user is read-only right?)

bin/run.sh
```diff
@@ -0,0 +1,40 @@
...
+
+case $1 in
+  version)
+    cat version.json
+    ;;
+  lambda.zip)
+    cd .venv/lib/python3.6/site-packages/
+    zip -r ../../../../lambda.zip *
+    ;;
+  kinto)
```
Natim: This is confusing, is it suppose to be the action we use to run kinto in production or just locally?
If we keep kinto-dist for production, aren't we suppose to provide elasticsearch as well (kinto-elasticsearch)?

TBH, I would leave kinto and initialize-kinto out of the scope of that docker image to prevent confusion with kinto-dist.

circle.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    # build the actual deployment container
+    - docker build --pull -t app .
+
+    # Clean up any old images; save the new one.
+    - I="image-$(date +%j).gz"; mkdir -p ~/docker; rm ~/docker/*; docker save app | pigz --fast -c > ~/docker/$I; ls -l ~/docker
+
+test:
+  override:
+    -  "true"
```
Natim: Do we have buildhub tests that we could run here to make sure the image is working?

Natim: > I didn't setup the unit tests task in the container because I was running into permissions issues (the app user is read-only right?)

I ran into the same issue for PollBot but I was able to write in /tmp

bin/run.sh
leplatrem: Yeah, I agree. 

Let's keep `initialize-kinto` though since it takes the `initialization.yml` as default parameter

Dockerfile
```diff
@@ -1,8 +1,28 @@
-FROM python:3.6
+FROM python:3.6-slim
+MAINTAINER Product Delivery irc://irc.mozilla.org/#product-delivery
+
+ENV KINTO_INI /etc/kinto/kinto.ini
+ENV PORT 8888
```
mostlygeek: use port 8000. It's more of a convention in ops.

leplatrem:
I believe we can merge, what do you think @Natim ?

docs/jobs.rst
```diff
@@ -106,6 +108,34 @@ This will produce a zip file that has to be uploaded in AWS Lambda configuration
...
+The exhaustive list of available commands and description is available using:
+
+::
+
+    docker run -t mozilla/buildhub
+
+For example, run a Kinto server:
+
+::
+    docker run -p 8888:8888 -t mozilla/buildhub kinto
```
Natim: I believe this is not the case anymore

PR #249, opened 2017-09-14
Add account:lambda to bucket write permissions

leplatrem:
That would be useful, only if the `account:lambda` is not the one used to initialize the server :)


jobs/buildhub/initialization.yml
```diff
@@ -2,6 +2,8 @@ build-hub:
   permissions:
     read:
       - system.Everyone
+    write:
+      - account:lambda
```
Natim: Actually this is not necessary because the user running the command will become an admin automatically. If we run it with the lambda user then it is already the case.

leplatrem: Agree.

PR #283, opened 2017-10-10
Add ability to configure cache folder via environment variable CACHE_FOLDER

leplatrem:
Untested locally


Dockerfile
```diff
@@ -18,6 +18,13 @@ RUN \
 RUN groupadd -g 10001 app && \
     useradd -M -u 10001 -g 10001 -G app -d /app -s /sbin/nologin app
 
+RUN \
+    mkdir /app/cache && \
+    chgrp app /app/cache && \
+    chmod g+w /app/cache
```
Natim: u+w

leplatrem: I did `chgrp`
I can change to `chown`

leplatrem: We should probably do it the other way and use `$CACHE_FOLDER` :)

PR #292, opened 2017-10-11
Keep trace but skip build urls that have unsupported formats

leplatrem:
@Natim I can't figure out why the `mock` has no effect...

### DIFF
--- a/jobs/tests/test_inventory_to_records.py
+++ b/jobs/tests/test_inventory_to_records.py
@@ -1,5 +1,6 @@
 import asyncio
 import json
+from unittest import mock
 
 import aiohttp
 import asynctest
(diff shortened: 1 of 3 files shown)


jobs/tests/test_inventory_to_records.py
```diff
@@ -634,3 +635,12 @@ def tearDown(self):
                 }
             }
         }
+
+    @mock.patch('buildhub.utils.guess_mimetype',
```
Natim: Try to mock guess_mimetype directly in the inventory_to_record file.

leplatrem: `guess_mimetype` is only used in `utils`

PR #297, opened 2017-10-27
Fix missing metadata for nightly archives (Fixes #296)

### LINKED ISSUE #296: S3 event Lambda fails when nightly archive is published before JSON
screenshot from 2017-10-23 16-35-34

leplatrem:
The problem was due to the fact that our tests didn't return a 404 HTML response but a connection refused when the URL was not mocked by `aiohttpresponses`

Fixes #296


leplatrem:
I thought I could reproduce the bug, but I can't this morning (neither did Travis with those changes, since they should fail)

PR #308, opened 2017-11-24
Get rid of dev instance

leplatrem:
@Natim  can you remove `/dev/` from UI in the hairy `deploy` command from `package.json` (I spent 10min on it and gave up...)

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -17,7 +17,7 @@
   },
   "homepage": "https://mozilla-services.github.io/buildhub",
   "scripts": {
-    "deploy": "npm run build && rimraf tmp && mkdir tmp && ls build/static/js/main* build/index.html | xargs sed -i 's|/buildhub/|/buildhub/stage/|g' && cp -R build/ tmp/stage/ && ls build/static/js/main* | xargs sed -i 's/buildhub.stage.mozaws.net/kinto-ota.dev.mozaws.net/' && ls build/static/js/main* build/index.html | xargs sed -i 's|/buildhub/stage/|/buildhub/|g' && cp -R build/* tmp/ && gh-pages -d tmp --add && rimraf tmp",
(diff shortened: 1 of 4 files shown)


ui/package.json
```diff
@@ -17,7 +17,7 @@
   },
   "homepage": "https://mozilla-services.github.io/buildhub",
   "scripts": {
-    "deploy": "npm run build && rimraf tmp && mkdir tmp && ls build/static/js/main* build/index.html | xargs sed -i 's|/buildhub/|/buildhub/stage/|g' && cp -R build/ tmp/stage/ && ls build/static/js/main* | xargs sed -i 's/buildhub.stage.mozaws.net/kinto-ota.dev.mozaws.net/' && ls build/static/js/main* build/index.html | xargs sed -i 's|/buildhub/stage/|/buildhub/dev/|g' && cp -R build/ tmp/dev/ && ls build/static/js/main* | xargs sed -i 's/kinto-ota.dev.mozaws.net/buildhub.prod.mozaws.net/' && ls build/static/js/main* build/index.html | xargs sed -i 's|/buildhub/dev/|/buildhub/|g' && cp -R build/* tmp/ && gh-pages -d tmp --add && rimraf tmp",
+    "deploy": "npm run build && rimraf tmp && mkdir tmp && ls build/static/js/main* build/index.html | xargs sed -i 's|/buildhub/|/buildhub/stage/|g' && cp -R build/ tmp/stage/ && ls build/static/js/main* | xargs sed -i 's/buildhub.stage.mozaws.net/kinto-ota.dev.mozaws.net/' && ls build/static/js/main* build/index.html | xargs sed -i 's|/buildhub/stage/|/buildhub/|g' && cp -R build/* tmp/ && gh-pages -d tmp --add && rimraf tmp",
```
leplatrem: instead of replacing in `build/` from prod to stage, and from stage to prod, wouldn't it be simpler to replace in `tmp/stage/` directly?

Natim: Today the default server in the source code is the stage endpoint, do you want me to default it to prod?

leplatrem: plus, why is there still a mention of `/kinto-ota.dev.mozaws.net`? 

let me try now that I understand what it does

leplatrem: @Natim OK I had realized that stage was the default. I switched to prod :)

PR #311, opened 2017-12-01
Fix querystring example for build ids

leplatrem:
@Natim @calixteman r?

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -158,7 +158,7 @@ In order to check that a build id exists, we'll just check that it is mentioned 
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -158,7 +158,7 @@ In order to check that a build id exists, we'll just check that it is mentioned
 
 .. code-block:: bash
 
-    curl -s $SERVER/buckets/build-hub/collections/releases/search?q=build.id=20170713200529 | \
+    curl -s $SERVER/buckets/build-hub/collections/releases/search?q=build.id="20170713200529" | \
```
Natim: Actually I don't think it is the case for the search endpoint is it?

leplatrem: It does, and it's less confusing I think

leplatrem:
@Natim r?

Issue #319, opened 2018-01-05
webui is missing links to repo, issue tracker, and docs


willkg:
The webui is missing some key bits which would make it easier for other people to collaborate. When looking at the webui, there's no indication of any of the following:

1. Where I can see the code.
2. Where I can file an issue.
3. How to file issues about data problems vs. code problems. Is it the same process? Is there anything that helps you triage/debug/work on things more effectively?
4. Copyright/license information for the data collected. I'm guessing it's public domain, but I have no idea.
5. What I can do with the data.
6. Link to documentation for the webui and the API.

I think the best thing to do is add a footer or a header with a link to the GitHub repository with "Source code", a link to the issue tracker "File an issue here", and a link to the readthedocs docs with "Documentation".

leplatrem:
Thanks for this feedback, indeed that's easy to fix and I agree that it'd have many benefits

peterbe:
This is the kind of questions the `/contribute.json` can answer. But a footer is also a good idea. 
One curious thing is, since buildhub is only available as a "sub directory" of `mozilla-services.github.com` it's not obvious what the URL ought to be. But I guess the only logical URL would be `https://mozilla-services.github.io/buildhub/contribute.json`

willkg:
`contribute.json` is really for machines--not humans. Adding one is great, but doesn't solve my problem. I'm asking for links in a header or footer or menu or something that explain the site, point to docs, point to support, docs on how to raise issues, etc.

PR #324, opened 2018-01-10
contribute.json

peterbe:
This is just a start to tackle https://github.com/mozilla-services/buildhub/issues/319

Once this is reviewed, we need to get it into `gh-pages` somehow.

### DIFF
--- a/contribute.json
+++ b/contribute.json
@@ -1,0 +1,20 @@
+{
+    "name": "buildhub",
+    "description": "Firefox Build Metadata Service",
+    "participate": {
+        "irc": "irc://irc.mozilla.org/#storage-team",
+        "irc-contacts": ["leplatrem", "magopian", "mostlygeek"],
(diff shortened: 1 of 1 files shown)


contribute.json
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "buildhub",
+    "description": "Firefox Build Metadata Service",
```
leplatrem: I would put *Mozilla Product* instead of just Firefox

contribute.json
```diff
@@ -0,0 +1,20 @@
...
+    "participate": {
+        "irc": "irc://irc.mozilla.org/#storage-team",
```
leplatrem: we have a `#product-delivery` channel, I think we should agree on using only one /cc @mostlygeek

peterbe: So what should I put?

leplatrem: Leave it like this I guess

peterbe: left it as is :)

PR #337, opened 2018-02-13
Try to avoid more non-manifest-containing subdirectories

glasserc:
Try to address https://bugzilla.mozilla.org/show_bug.cgi?id=1437931 by skipping anything that doesn't look exactly like a date.


glasserc:
This is just a slapped-together attempt to speed a resolution -- feel free not to use it or to discard it if you don't think it's helpful.

Issue #338, opened 2018-02-14
*.zip and *.installer.exe are mixed


calixteman:
here: https://mozilla-services.github.io/buildhub/?platform[0]=win64&channel[0]=nightly&locale[0]=en-US&products[0]=firefox
we can see that for win32 and win64, .zip and .installer.exe are mixed (it's visible on column "Size").

leplatrem:
We can't know in advance which archive will come first on S3.
I would be tempted to believe that the second one overwrites the record coming from the first archive.

Would you have a preference? Looks like .exe is everywhere whereas zip are just for nightly no?

Issue #318, opened 2018-01-05
page showing builds by buildid in webui


willkg:
It'd be great if the webui had a page that showed all the builds/build-info associated with a given buildid.

For example, maybe a url like this?:

https://mozilla-services.github.io/buildhub/?buildid[0]=20180105100137

(I'm imitating the url structure without really understanding what it means.)

Then that page would show the information filtered by that buildid.

The reason I bring this up is that it'd help in Socorro to link buildids in our webapp to some canonical url with information about that build. Buildhub seems like the right choice currently.

The Socorro bug is https://bugzilla.mozilla.org/show_bug.cgi?id=687630 .

leplatrem:
I think it's already possible using `?q=`: https://mozilla-services.github.io/buildhub/?q=20180105100137 

Do you think that would be enough and suit your needs?

willkg:
I think I'd prefer a `buildid` filter rather than just doing a general query because the url is clearer, it's more intentional, and codifies that functionality in the buildhub app.

I think using a general query would work for now. I'll add links from Socorro using the general query link and we'll see if any issues come up and deal with them then.

Thanks!

PR #340, opened 2018-02-15
Archive link on file size column (ref #338)

leplatrem:
ref #338 

I find this more consistent than on the version column


leplatrem:
> But now there's no way at all to get to the URL?

Yes, on the size column instead of the version...

> I too find it confusing that for each version there's more than one.

That's a broader issue (related to #318 somehow)

peterbe:
> Yes, on the size column instead of the version...

Grr! I'm blind.

PR #342, opened 2018-02-15
Retry requests on 409 responses

leplatrem:


glasserc:
I don't love this because I don't feel like we understand why we get 409s, but this is also just a standard dependency bump so let's do it.

PR #343, opened 2018-02-15
loading contribute.json

peterbe:
Same thing but without needing an XHR. 
The `/contribute.json` URL still doesn't work but I'm not too worried yet because buildhub is likely to be moved into delivery console some day soon.


leplatrem:
> The /contribute.json URL still doesn't work

It does once moved to the `public/` folder

PR #351, opened 2018-02-21
Now pick Windows .exe archives only (fixes #338)

### LINKED ISSUE #338: *.zip and *.installer.exe are mixed

leplatrem:
Fixes #338


jobs/buildhub/utils.py
```diff
@@ -159,13 +159,18 @@ def is_build_url(product, url):
     if re_exclude.match(url):
         return False
 
+    # Only .exe for Windows.
+    extensions = list(FILE_EXTENSIONS)
+    if 'win' in url:
```
Natim: arf...

PR #352, opened 2018-02-22
Prepare release 1.1.5

leplatrem:


jobs/requirements.txt
```diff
@@ -21,4 +21,4 @@ six==1.11.0
 Unidecode==1.0.22
 urllib3==1.22
 wrapt==1.10.11
-yarl==1.1.0
+yarl==1.1.1
```
peterbe: Is this related to the 1.1.5 release?

jobs/setup.py
```diff
@@ -35,7 +35,7 @@ def read_file(filename):
 
 
 setup(name='buildhub',
-      version='1.1.5.dev0',
+      version='1.1.5',
```
peterbe: I get nervous around version stuff. Is `1.1.5 > 1.1.5.dev0`?

jobs/requirements.txt
leplatrem: When releasing, we usually pin the dependencies to their latest version (running `make build-requirements`). If tests pass, we tag.

jobs/setup.py
leplatrem: Yes. BTW we use `zest.releaser` in most of our python projects.

jobs/requirements.txt
peterbe: Uh. I don't know if I get that. It sounds really dangerous. 
Upgrading versions should be slow and deliberate and ideally you should be running the new versions locally a bit before you decide to start a pull request, CI and eventually release.

leplatrem: To be honest, I almost never run the code locally :) Especially buildhub!

I get your point, but usually I'm confident enough in the test suite to be able to upgrade a minor version in the dependencies.

Besides that, I'm really scared when a project does not have [tests as specifications](http://blog.mathieu-leplatre.info/your-tests-as-your-specs.html) or a low test code coverage.

Natim: > Uh. I don't know if I get that. It sounds really dangerous.

Actually here we need to explain that by default when you install the app it always use the last version, so at release time we take the current set of working dependencies to mark it in time so that we can use always the same set in production.

I understand your concerns and I guess moving to pipenv would be a better way of handling this.

I am an optimism and I like to think that upgrading dependencies automatically is a way to keep a project uptodate with noticing update errors only when the CI is failing.

All our projects set their dependencies in the `setup.py` and try to run using the last version of the dependencies (which works most of the time) we have a cronjob running in the CI that will detect when it is not the case anymore and in that case we create an issue to address it.

At time of release we make a list of all the dependencies versions so that we can constraint pip to use this dependencies version while installing:

    pip install kinto -c requirements.txt

See https://pip.pypa.io/en/stable/user_guide/#constraints-files

PR #356, opened 2018-02-23
pin all requirements, fixes #353

### LINKED ISSUE #353: Pin all python requirements
We're currently running without pinned shas in [the requirements](https://github.com/mozilla-services/buildhub/blob/0fa1fc0d917f63d47fa7f2586a6e495ddb4c5ad3/jobs/requirements.txt). We should take this opportunity to use default/constraints to split it up.

peterbe:
Just getting started. A lot to learn still. 

What this humble start does is that it moves all package requirements into two files.
I can at least now do:
```
$ pip install -r jobs/requirements/default.txt -c jobs/requirements/constraints.txt
```
and that install all the packages that once was un-pinned in `setup.py`.

What's next?

- [x] Need to figure out how Travis installs things
- [x] Rewrite the Release instructions to not depend on `pip install -e .` etc. 
- [x] Both travis and circle needs to run the `pip install ...` command mentioned above. 
- [x] Change the docker scripts that generates the virtualenv and makes a zip out of it.


leplatrem:
What do you think about using ` pipenv`? It provides default/dev version sets in Pipfile.lock, hashes etc.

jobs/setup.py
```diff
@@ -57,5 +48,8 @@ def read_file(filename):
       package_data={'buildhub': ['initialization.yml']},
       include_package_data=True,
       zip_safe=False,
-      install_requires=REQUIREMENTS,
+      # Use
+      #  `pip -r requirements/default.txt -c requirements/constraints.txt`
+      # instead.
+      install_requires=[],
```
leplatrem: I find that rather unsual, do you have many examples of projects doing that? And that prevents using the package anywhere else, right?

Natim: This is the django way of handling things and it works until your projects become a dependency on another project. (Which is the case with django project but not django libs)

leplatrem: > This is the django way of handling things

Ooh dear... Django...

I'm even tempted to believe that `setup.py` is useless without `install_requires=[]`.

And yes you're right, I remember, at the time some people did not manage their Django projects as Python eggs, and had just a requirements.txt file. I never liked it :smile: 

Anyway, I think we can do better than that nowadays ;)

peterbe:
To be clear; what I want is to make things more predictable and easier to maintain. One huge step for that goal is to make absolutely sure that the exact same versions of stuff you have locally when working deep on something is the exact same versions you get in CI and in production. It's hard enough to debug things considering that nobody can ever have AWS S3 or TaskCluster in their laptops. 

The goal of this is to remove any doubt that you have the exact same stack locally as you have in CI and production.

jobs/setup.py
peterbe: > I find that rather unsual, do you have many examples of projects doing that?

[Tecken](https://github.com/mozilla-services/tecken/blob/master/setup.py) 
[Socorro](https://github.com/mozilla-services/socorro/blob/master/setup.py#L42)
[Telemetry Analysis Service](https://github.com/mozilla/telemetry-analysis-service/blob/master/setup.py)

By the way, all of these above projects are 100% Docker, which all (I think) test and build in CI then get shipped as a docker container into EC2 servers so the CI build is extremely close to the EC2 runtime servers.

peterbe: > And that prevents using the package anywhere else, right?

Indeed. This whole technique is only for "deployment projects". 
The origins from this is that there was no way of doing `pip install -e .` (or `python setup.py develop`) AND get the checksum integrity that only peep.py could do. `peep.py` was later merged into pip core around version 9 (?).
But it's no harm not being able to do `pip install -e .` when you can easily swap that for `pip install -r requirements.txt` and get all the predictability and security that checksums offers. 

By the way, I think when Erik built `peep.py` he had in mind that developers should audit the installed package's source code to look for backdoors and if you establish that there are no backdoors locally, you won't have any backdoors in production. I honestly don't think anybody does this audit check. I certainly don't.

peterbe:
> What do you think about using `pipenv`?

I have to admit that I'm way behind on `pipenv` and should maybe go learn more about it for production systems.

jobs/setup.py
leplatrem: See for Socorro: https://github.com/mozilla-services/socorro/pull/2516  and https://bugzilla.mozilla.org/show_bug.cgi?id=1110521

peterbe:
I'm still a [pipenv](https://docs.pipenv.org/) newbie but it seems attractive for us:

* You don't have to ask developers to first create a virtualenv and then install dependencies. 
* It automatically splits up the packages you mention (and care about) away from the tedious task of having to list all the sub-dependencies.
* Still doesn't make `python setup.py install` work and taking care of installing dependencies. But not sure we ever need that in this scope. 
* You separate runtime vs. dev dependencies with the `--dev` flag. In something like tox, CI and on devs' laptops you run `pipenv install --dev` but for production deployment you use `pipenv install` and it won't bother with `py.test` etc. 

Things to keep in mind (aka. friction):

* [Pipfiles support in PyUP.io](https://github.com/pyupio/pyup/issues/197) is still an open issue but it mentions that it "should" work even though it's not officially completed. However, it says [*"The bot won't update the hash in Pipfile.lock."*](https://github.com/pyupio/pyup/issues/197#issuecomment-360391226) which means it's still useless.
* We have to change from virtualenv and regular virtualenv shell activation to using `pipenv`'s way of doing things. This might add complexity that is out of scope from the goal of "pinning requirements more hard". 


Depending on how this go I'm going to try to get this PR to at least work WITHOUT `pipenv` and then either add more effort or start a new issue to switch to `pipenv`. Gut-feeling for the time of writing.

peterbe:
Actually, [been thinking about it some more](https://github.com/mozilla-services/socorro/pull/4333#issuecomment-368543736) and considering that we probably will enable Pyup for `buildhub` and since Pyup [doesn't support `Pipefile`](https://github.com/pyupio/pyup/issues/197) yet I don't think we should use it. 

Combine that argument with that fact that this PR tries to just harden dependencies, switching to `pipenv` should definitely wait.

Dockerfile
```diff
@@ -12,8 +12,7 @@ WORKDIR /app
 
 RUN \
     make virtualenv && \
-    .venv/bin/pip install --constraint jobs/requirements.txt jobs/ && \
-    .venv/bin/pip freeze | grep -v -- 'buildhub' > dependencies.txt
+    .venv/bin/pip install -r jobs/requirements/default.txt -c jobs/requirements/constraints.txt
```
peterbe: I still don't really get this line. What's the point?

If the point is to make a virtualenv *only* for the benefit on `make lambda.zip` then why can't this be run as part of that? Only.

README.md
```diff
@@ -2,7 +2,7 @@
 
 [![Build Status](https://travis-ci.org/mozilla-services/buildhub.svg?branch=master)](https://travis-ci.org/mozilla-services/buildhub)
 
-*Buildhub* aims to provide a public database of comprehensive information about releases and builds.
+_Buildhub_ aims to provide a public database of comprehensive information about releases and builds.
```
peterbe: Sorry. My editor uses Prettier on every save by default. Apparently this is the Prettier way of doing Markdown.

bin/run.sh
```diff
@@ -38,6 +38,6 @@ case $1 in
     latest-inventory-to-kinto ${@:2}
     ;;
   *)
-    usage
+    exec "$@"
```
peterbe: This is nifty. Any command that you run after `docker run  -it mozilla/buildhub` would be basically executed. E.g. `docker run  -it mozilla/buildhub bash` or `docker run  -it mozilla/buildhub ls /tmp`

jobs/tox.ini
```diff
@@ -5,9 +5,13 @@ skip_missing_interpreters = True
...
+    python setup.py develop
     py.test --cov-report term-missing --cov buildhub tests {posargs}
+
 deps =
-    -rdev-requirements.txt
+    -r{toxinidir}/requirements/dev.txt
+    -r{toxinidir}/requirements/default.txt
+    -c{toxinidir}/requirements/constraints.txt
 
 install_command = pip install --pre {opts} {packages}
```
peterbe: I don't think we need the `--pre` any more since we're not allowing installation of anything else except exactly what we're prepared to install.

jobs/tox.ini
```diff
@@ -5,9 +5,13 @@ skip_missing_interpreters = True
 [testenv]
 usedevelop=true
 commands =
+    python setup.py develop
```
peterbe: In Docker, to install `buildhub` itself, I used `pip install -e jobs` which effectively executes the `setup.py`. But I don't know how to do that in tox.

peterbe:
For the record this worked:

```sh
$ make lambda.zip
```
It created a `buildhub-lambda-v1.1.5-6-g2f1da14.zip` which when I inspect contains a bunch of `.pyc` files like `wrapt/__pycache__/decorators.cpython-36.pyc`. Is that really the right thing to install in AWS Lambda?

Dockerfile
leplatrem: To be honest I don't remember why we generate this `dependencies.txt` file.

jobs/tox.ini
leplatrem: I never was a tox expert, and can't help much :( But I can't remember having to do this though

jobs/tox.ini
leplatrem: Yes indeed

jobs/tox.ini
Natim: With tox your can use `usedevelop = true`

PR #358, opened 2018-02-27
circle only

peterbe:
The goal is to have *only* the `circle.yml`. No `tox.ini` and no `.travis.yml`.

- [x] Run flake8 in circle in docker
- [x] Run functional test in docker
- [x] Delete `tox.ini`
- [x] Delete `.travis.yml`
- [x] Build `lambda.zip` (and `buildhub-lambda-v1.x.x.x-SHA.zip`) in docker
- [x] upgrade to circleci 2.0
- [x] Circle should run `npm run build` in the `ui` directory too
- [x] Generate lambda.zip in Circle and check that it doesn't barf
- [x] Check that lambda.zip does contain `buildhub/*.pyc` files and not just an egg-link.
- [x] Move flake8 configs (used to be in `tox.ini`) into `setup.cfg` or `pytest.init`
- [x] All `ui` related stuff in Circle should use Docker (and/or docker-compose)
- [x] Do what the old `circle.yml` file did with dockerhub pushing.
- [x] Build docs in CI to make sure it compiles at all
- [x] Clean up (or remove) Makefile
- [x] Run lintcheck of ui code in CI too


bin/run.sh
```diff
@@ -20,14 +20,19 @@ case $1 in
     cat version.json
     ;;
   test)
-    python3 -m venv /tmp/tests
-    /tmp/tests/bin/pip install -e jobs/
-    /tmp/tests/bin/pip install -r jobs/requirements/default.txt -r jobs/requirements/dev.txt -c jobs/requirements/constraints.txt
-    /tmp/tests/bin/py.test --ignore=jobs/tests/test_lamdba_s3_event_functional.py --override-ini="cache_dir=/tmp/tests" jobs/tests
+    py.test --ignore=jobs/tests/test_lamdba_s3_event_functional.py --override-ini="cache_dir=/tmp/tests" jobs/tests
```
peterbe: This uses the `py.test` installed in Docker only.

bin/run.sh
```diff
@@ -20,14 +20,19 @@ case $1 in
...
+    rm -fr .venv
+    virtualenv .venv
+    source .venv/bin/activate
+    pip install -e jobs
+    pip install -r jobs/requirements/default.txt -c jobs/requirements/constraints.txt
+    pushd .venv/lib/python3.6/site-packages/
     zip -r /tmp/lambda.zip *
+    popd
+    cp /tmp/lambda.zip .
     ;;
```
peterbe: This makes sure the creation of the `lambda.zip` is exclusive. It only has `buildhub` and `requirements/default.txt` in a dedicated virtualenv that exists purely for the sake of being packaged into a `.zip`.

One thing that makes me nervous about this is the fact that `buildhub` itself is not in the .zip. 

```sh
▶ aunpack -l buildhub-lambda-v1.1.5-3-g3d56a0c.zip | grep buildhub
Archive:  buildhub-lambda-v1.1.5-3-g3d56a0c.zip
       11  02-27-2018 15:52   buildhub.egg-link
```
Is that correct @leplatrem ?

peterbe:
@leplatrem What creates this `.metadata-1.1.5.json` file? And should it be git ignored?

bin/run.sh
leplatrem: Hmm, maybe not, look at the latest release:

```
unzip -l buildhub-lambda-v1.1.5.zip | grep buildhub
Archive:  buildhub-lambda-v1.1.5.zip
        0  2018-02-23 15:34   buildhub/
     8292  2017-10-11 11:18   buildhub/to_kinto.py
    16121  2018-02-22 15:51   buildhub/inventory_to_records.py
    14247  2018-02-22 15:51   buildhub/utils.py
     6584  2017-09-14 14:24   buildhub/initialization.yml
     6953  2018-02-15 09:52   buildhub/lambda_s3_inventory.py
        0  2017-05-26 15:35   buildhub/__init__.py
[...]
```

leplatrem:
> @leplatrem What creates this .metadata-1.1.5.json file? 

Buildhub has two main «jobs». One is called via a S3 event when a file is created, and another is executed weekly in a cronjob, from the exhaustive S3 inventory files, mainly in order to catch up with the potential missing archives.
For the inventory job, we reuse the fetched metadata from one to another. 

https://github.com/mozilla-services/buildhub/blob/53518a873b45daf497347ac71770859ec6ce2b8a/jobs/buildhub/inventory_to_records.py#L346-L353

> And should it be git ignored?

Yes

Makefile
```diff
@@ -6,30 +6,38 @@ SPHINX_BUILDDIR = docs/_build
...
-virtualenv: $(PYTHON)
-$(PYTHON):
-	$(VIRTUALENV) $(VENV) --python=python3.6
+#
+# virtualenv: $(PYTHON)
+# $(PYTHON):
+# 	$(VIRTUALENV) $(VENV) --python=python3.6
+
+.docker-build:
+	make build
```
leplatrem: maybe this could depend on some source like setup.py to avoid having to run `make clean` manually

Dockerfile
```diff
@@ -1,30 +1,53 @@
...
+
+# Clean up apt
+RUN apt-get autoremove -y && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 COPY . /app
+
+# Install Python dependencies
+RUN pip install -e app/jobs/
```
leplatrem: As far as I understood your approach, this does not install Python dependencies at all (move comment below?)

peterbe: Correct. It just installs `buildhub` itself.

bin/run.sh
peterbe: I think it might be because of the use of `-e`. Thanks for pointing this out.

leplatrem:
:tada:

.circleci/config.yml
```diff
@@ -0,0 +1,161 @@
...
+          name: Lint check
+          command: |
+            docker-compose -f docker-compose.ci.yml run buildhub lintcheck
+
+      - run:
+          name: Mock run building lambda.zip.
+          command: |
+            docker-compose -f docker-compose.ci.yml run buildhub lambda.zip
+
+      - run:
```
peterbe: @bqbn Hey,
These below lines I took form [a working circle 2.0 project](https://github.com/mozilla-services/tecken/blob/master/.circleci/config.yml) that @milescrabill helped me build. 

Basically I just copied it. And replaced the name `tecken` with `buildhub`. 
See [the script too](https://github.com/peterbe/buildhub/blob/65910b885b47dd6bc4b5a873b2288a7cd0fb521c/bin/deploy-dockerhub.sh)

The build image name is specified in the [`docker-compose.ci.yml`](https://github.com/peterbe/buildhub/blob/65910b885b47dd6bc4b5a873b2288a7cd0fb521c/docker-compose.ci.yml#L21) as `mozilla/buildhub`. 

I appreciate that it's a bit big to review but if you could please take look at the lines specifically for the "Push to Dockerhub". 

I have no idea how to test any of this. And if we don't get it right here in this monster PR, we can admittedly iterate over it later.

bin/deploy-dockerhub.sh
```diff
@@ -0,0 +1,35 @@
...
+        if [[ $count -gt $max ]]; then
+            return 1
+        fi
+        echo "$count / $max"
+    done
+    return 0
+}
+
+# configure docker creds
+retry 3  echo "$DOCKER_PASSWORD" | docker login -u="$DOCKER_USERNAME" --password-stdin
```
bqbn: I suspect this only retries the "echo" command 3 times, not the "docker login" command.

.circleci/config.yml
bqbn: The below looks good to me. The way I would test it is just let CircleCI build it and see if what errors come out of it. If success, then the log should show it successfully pushed an image to dockerhub.

peterbe:
@leplatrem It's ready for review. There are things that (might) need work. For example, some of the docs are still referring to old ways of doing things. I guess I just got overwhelmed and wanted some peer feedback before I proceed editing that too. 
I still don't know of a way to simulate or test S3 Events locally. Perhaps that's something we can discuss later. Who knows, we *might* do things differently anyway once we have the `buildhub.json` files to scrape.

peterbe:
@magopian @glasserc I don't know how much experience you have on the Buildhub non-ui side of things but if you have time, could you take a look at this PR from a "high level". I.e. take a look at the dev.rst and see if you can run the various things and if the instructions make sense.

docs/api.rst
```diff
@@ -17,7 +17,7 @@ Servers
 Clients
 =======
 
-Since it is an HTTP API and the records are public, you can simply use any HTTP client, like `curl <http://curl.haxx.se>`_ or `HTTPie <https://httpie.org>`_.
+Since it is an HTTP API and the records are public, you can simply use any HTTP client, like `curl <http://curl.haxx.se>`_ or `HTTPie <https://httpie.org>`__.
```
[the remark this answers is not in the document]
peterbe: Without it you get an warning from `sphinx` and with the double underscore the warning goes away. 
You get this:
```
/app/docs/api.rst:4: WARNING: Duplicate explicit target name: "httpie".
```
See https://stackoverflow.com/questions/5464627/how-to-have-same-text-in-two-links-with-restructured-text/14067756#14067756

It's because there are two hyperlinks with the same text but different URL. 

The two different urls were `https://httpie.org` and `https://httpie.org/`. Subtle. 
As a matter of fact, if I correct that so both URLs are `https://httpie.org` instead, then I can use a single underscore instead.

docs/dev.rst
```diff
@@ -0,0 +1,214 @@
...
+    app@b95573edb130:~$ latest-inventory-to-kinto
+
+That'll take a while.
+
+Functional tests
+================
+
+The functional tests require that you have a ``kinto`` server up and running.
+By default, the functional tests assumes that the ``kinto`` server is running
+at ``http://testkinto:9999``. To start that server, run:
```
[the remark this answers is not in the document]
peterbe: Done. I wrote a little something.

PR #382, opened 2018-03-23
skip old csv entries, fixes #380

### LINKED ISSUE #380: Ability to limit by date in `csv_to_records`
When you scrape with `latest-inventory-to-records` it analyses *every single row* from the CSV files downloaded from the manifest in S3. 

This makes it exceptionally time consuming to run the whole thing. Not only is it inconvenient when running it multiple times (because might be testing something) but when so many operations have to be done, the risk of a bad network error somewhere blowing everything up is real. 

Alternatively, we could use this to do something like:

* Once a week/day run the whole thing in a cron job. Pretty sure we have all the firefox 20, 21, 22 etc. releases in the database, but if our algorithm changes or improves it'd be nice to revisit those old releases from S3 and try them again differently. 

* If it's now not so horribly slow, you can run the cron job every hour instead. That's more quickly patch runtime flaws in the Lambda executions.

peterbe:
This works locally. I ran it with 
```
NB_PARALLEL_REQUESTS=4 MIN_AGE_LAST_MODIFIED_HOURS=48 latest-inventory-to-kinto
```
and it only let through 508 entries to become records to send to kinto. 508 builds in the last 48 hours. 

Perhaps this is silly. The manifest is updated once a day by S3. If you run the (the cron job) `latest-inventory-to-kinto`, say, 5 times per day the manifest will point to exact same csv listings every single time. If it worked **once**, with `MIN_AGE_LAST_MODIFIED_HOURS=24` perhaps there's no point at all to run it ever again that day. 

Perhaps a more realistic approach is to use `MIN_AGE_LAST_MODIFIED_HOURS=168`, which is 7 days. On that note, perhaps it's more appropriate to call it `MIN_AGE_LAST_MODIFIED_DAYS` as the unit instead. Arguably the less the unit the more flexibility.


.gitignore
```diff
@@ -12,3 +12,4 @@ version.json
 buildhub-lambda-*.zip
 .docker-build
 .metadata-*.json
+.records-kinto-build-hub-releases.json
```
peterbe: I'm surprised this wasn't already in the `.gitignore`. It's technically out-of-scope for this PR but I snuck it in anyway. YOLO!

PR #384, opened 2018-03-27
80 lines pep8

peterbe:


leplatrem:
Just FYI we use 99 chars in other projects
https://github.com/Kinto/kinto/blob/6d77dcb2b94f9198fa08db489776440f7df0c275/tox.ini#L45-L46

PR #388, opened 2018-03-28
Add statsd metrics

peterbe:


jobs/buildhub/configure_markus.py
```diff
@@ -0,0 +1,41 @@
...
+        # http://markus.readthedocs.io/en/latest/usage.html#markus-configure
+        markus.configure([
+            # Deliberately commented out. Uncomment if you want to see
+            # what's sent to markus in local development.
+            # {
+            #     'class': 'markus.backends.logging.LoggingMetrics',
+            #     'options': {
+            #         'logger_name': 'metrics'
+            #     }
+            # },
```
leplatrem: we'd rather add a config variable to include/exclude this block instead of using code un/commenting

jobs/buildhub/lambda_s3_event.py
```diff
@@ -226,15 +229,17 @@
...
+                for record in records_to_create:
+                    # Check that fields values look OK.
+                    utils.check_record(record)
+                    # Push result to Kinto.
+                    kinto_client.create_record(data=record,
+                                               bucket=bucket,
+                                               collection=collection,
+                                               if_not_exists=True)
+                    logger.info('Created {}'.format(record['id']))
+                    metrics.incr('s3_event_record_created')
```
leplatrem: That's out of scope, but I'm surprised we don't use the `batch()` APi here :)

```py
with kinto_client.batch() as batch:
    for record in records_to_create:
        batch.create_record(....)
```

jobs/buildhub/configure_markus.py
peterbe: yeah, removed now. There's a URL in the comment to the documentation for markus.

jobs/buildhub/lambda_s3_event.py
peterbe: What are the chances of there being "many" records to create? If a new file is added in S3, the S3 Event is triggered, which triggers the Lambda function. So almost always there'll only be one record, right?

leplatrem: Let's say we have the following S3 events:
1. fr-FR archive, no metadata, no record.
1. en-US archive, no metadata, no record.
1. en-US metadata, scan of locale folders, 2 records.

PR #389, opened 2018-04-05
bail on 404 responses in cron job

peterbe:


jobs/buildhub/inventory_to_records.py
```diff
@@ -65,6 +74,16 @@
                 headers=headers,
                 timeout=None
             ) as response:
+                # Calling 'response.raise_for_status()' when the response
+                # status code is 404 will trigger a ClientResponseError
+                # which is part of the backoff (see decorator).
+                # *Sometimes* a 404 response is actually something that can
+                # happen due to the latency of a slow update in
+                # archive.mozilla.org so if you just back off and try again
+                # in a couple of minutes it will work.
```
leplatrem: This comment is redacted in a way that is not directly related to the condition just below. Maybe it should mention that raising JSONFileNotFound will exit without retry or something...

peterbe: I think I'll move it all to the doc string of the function instead.

peterbe:
uh? functional tests passed just fine locally. sigh.

peterbe:
Ping @leplatrem (just so you don't forget)

The functional tests don't fail locally. Perhaps you can take a look and give me some clues as to what to look for in this one.

leplatrem:
Some weird behaviours may happen when postgresql runs with `fsync=off`. That may be the case in Continuous Integration services. If we rely on Docker, it should not be a problem though...

Let's have a chat' to corner this beast

peterbe:
There were quite a few places where it leaked. Hopefully all fixed now. 

I think what happened was that it would actually literally try to do `GET http://archive.mozilla.org/pub/firefox/nightly/2016/05/2016-05-02-03-02''-07-mozilla-central/firefox-49.0a1.en-US.linux-i686.json` for real. Which 404s. Before we used to lump them together as `ClientError` exceptions and swallow them. 

You know, where we do things like `self.addCleanup(self.mockresponses.stop)` in the tests. I wish there was a way to do something like this:
```python
def setUp(self):
    self.mockresponses = aioresponses(raise_on_leaked_urls=True)  # <-- this!
    self.mockresponses.start()
    
    self.mockresponses.get(URL, payload, status=whatever)
    ...

    self.addCleanup(self.mockresponses.stop)
```
Then, when it calls `self.mockeresponses.stop()` in the cleanup, it would be cool if it could compare the list of URLs used by the mocking and lists of URLs that flew through. 
Do you know if that's possible?

peterbe:
I'm going to test if we can use something like `pytest-blockade` (or whatever it's called) which I've never tested in an asyncio world. 

By the way, I noticed [this uglieness](https://github.com/mozilla-services/tecken/blob/9ce44f2f8a0a947d1988bcd93fb6ea873286e20c/pytest.ini#L9-L11) :) which I'm totally responsible for.

jobs/tests/test_lamdba_s3_event.py
```diff
@@ -65,6 +65,13 @@ def setUp(self):
         self.mockresponses.start()
         for url, payload in self.remote_content.items():
             self.mockresponses.get(utils.ARCHIVE_URL + url, payload=payload)
+        # This will be attempted every time since the metadata cache_folder
+        # is always reset by the pytest fixture.
+        self.mockresponses.get(
+            utils.ARCHIVE_URL + 'pub/firefox/nightly/2016/05/2016-05-02-03-02'
+            '-07-mozilla-central/firefox-49.0a1.en-US.linux-i686.json',
+            status=404
+        )
```
leplatrem: I find weird to have this here honestly.  For `CSVToRecords` it seems ok, because the remote_content is defined on the same class. But here, it's the parent class, I think we can do better :) 

What about:

```py
    for url, payload in self.remote_content.items():
        if instanceof(payload, int):
            self.mockresponses.get(utils.ARCHIVE_URL + url, status=payload)
        else:
            self.mockresponses.get(utils.ARCHIVE_URL + url, payload=payload)
```

And put this `firefox-49.0a1.en-US.linux-i686` where it belongs?

leplatrem:
> Then, when it calls self.mockeresponses.stop() in the cleanup, it would be cool if it could compare the list of URLs used by the mocking and lists of URLs that flew through.
> Do you know if that's possible?

That would be a nice feature! The code of https://github.com/pnuckowski/aioresponses is rather small, I believe it should not be super hard to implement. This is the code of the mocked aiohttp request:
https://github.com/pnuckowski/aioresponses/blob/e9ae2244043116ebde3d8ee45c98879f88edaedf/aioresponses/core.py#L205-L226

jobs/tests/test_lamdba_s3_event.py
peterbe: You're right! It's clear now when I "zoom out".

peterbe:
You know, I'm going to test something. Once this lands. 

The `utils.ARCHIVE_URL` is currently hardcoded to `https://archive.mozilla.org`. If I run tests like this:  `ARCHIVE_URL=https://archive.example.whatever pytest` it shouldn't fail with measly 404 errors but connection errors and that would be a way to spot leaking networking tests. 

In the Socorro Django tests we had a rule that you can never mention "Firefox" or "Thunderbird" in any of the test code, e.g. fixtures. It should always be "WaterWolf" or "NightTrain". Saved us a couple of times.

PR #403, opened 2018-04-13
Disk cache the manifests, fixes #392

### LINKED ISSUE #392: Disk cache the manifests
Every 24h about 80+% of the manifest sub-URLs change. It's the *.csv.gz files. 

But for 24h the files are 100% the same every time (since they have hashes in them). So if you download more than once within 24h you could greatly benefit for a disk cache. 

The production server that runs the cron job sits near S3 so the downloads are pretty fast. Also, we're only running the cron job every 24h at the moment. 

However, for local development it's near impossible to run the scraping more than once. It's just too much network downloading. Makes it hard to try and try again. 

Also, if we could implement a caching for every 24h we would be able to run the scraping cron job in production every 1h instead. That means we're much better equipped to recover if the cron fails. I.e. if the lambda event is missed and the scraping bombs out due you have to wait a whole other day to get it right. That definitely has happened in production.

peterbe:
The gist of this is that when we download the `.csv.gz` files we break it up.

1) First clear out any old downloaded `.csv.gz` files that are older than 48 hours. 
2) For each `.csv.gz` file in the manifest.json, instead of downloading it and yielding each CSV row, we download it to `$CSV_DOWNLOAD_DIRECTORY` (*)
3) After the file has been downloaded, instead of reading it from a network URL, we read it from the local file. 

(*) In docker-compose this becomes `./csv-download-directory` but outside Docker it becomes `$TMPDIR/csv-download-directory`. This way it's kept between stopping and starting Docker.


jobs/buildhub/s3_inventory_to_kinto.py
```diff
@@ -129,25 +139,79 @@ def ends_with_date(prefix):
...
-    async for key in keys_stream:
-        key = 'public/' + key
-        logger.info('Fetching inventory piece {}'.format(key))
-        file_csv_gz = await s3_client.get_object(Bucket=BUCKET, Key=key)
+
+    # Make sure the directory exists if it wasn't already created.
+    if not os.path.isdir(download_directory):
+        os.mkdir(download_directory)
+
+    print("LOOKING AT", download_directory)
```
peterbe: Oops. I have to delete this.

jobs/buildhub/s3_inventory_to_kinto.py
```diff
@@ -129,25 +139,77 @@ def ends_with_date(prefix):
...
     the csv.gz manifests.
     """
-    async for key in keys_stream:
-        key = 'public/' + key
-        logger.info('Fetching inventory piece {}'.format(key))
-        file_csv_gz = await s3_client.get_object(Bucket=BUCKET, Key=key)
+
+    # Make sure the directory exists if it wasn't already created.
+    if not os.path.isdir(download_directory):
+        os.mkdir(download_directory)
```
leplatrem: nit: use `makedirs(..., exists_ok=True)` ?

jobs/buildhub/s3_inventory_to_kinto.py
```diff
@@ -129,25 +139,77 @@ def ends_with_date(prefix):
...
+
+    # Look for old download junk in the download directory.
+    too_old = 60 * 60 * 24 * 2  # two days
```
leplatrem: move to `DOWNLOAD_MAX_AGE` constant ?

jobs/buildhub/s3_inventory_to_kinto.py
```diff
@@ -129,25 +139,77 @@ def ends_with_date(prefix):
...
+            os.remove(file_path)
+
+    async for files in files_stream:
+        # If it doesn't exist on disk, download to disk.
+        file_path = os.path.join(
+            download_directory,
+            files['MD5checksum'] + '.csv.gz'
+        )
+        # The file neither exists or has data.
+        if os.path.isfile(file_path) and os.stat(file_path).st_size:
```
leplatrem: nit:  add explicit `st_size > 0`

jobs/buildhub/s3_inventory_to_kinto.py
```diff
@@ -129,25 +139,77 @@ def ends_with_date(prefix):
...
+            key = 'public/' + files['key']
+            logger.info('Fetching inventory piece {}'.format(key))
+            file_csv_gz = await s3_client.get_object(Bucket=BUCKET, Key=key)
+            async with aiofiles.open(file_path, 'wb') as destination:
+                async with file_csv_gz['Body'] as source:
+                    while 'there are chunks to read':
+                        gzip_chunk = await source.read(chunk_size)
+                        if not gzip_chunk:
+                            break  # End of response.
+                        await destination.write(gzip_chunk)
```
leplatrem: Here maybe you should try/except and delete the partially downloaded file when an error happens, or maybe you don't want to check the md5sum when resuming

jobs/buildhub/s3_inventory_to_kinto.py
peterbe: Out-of-curiousity; why? 

Isn't it explicit already that the test is for the `st_size` to be anything greater than 0?

leplatrem: Matter of taste maybe :)

jobs/buildhub/s3_inventory_to_kinto.py
```diff
@@ -129,25 +141,82 @@ def ends_with_date(prefix):
...
+                            gzip_chunk = await source.read(chunk_size)
+                            if not gzip_chunk:
+                                break  # End of response.
+                            await destination.write(gzip_chunk)
+                size = os.stat(file_path).st_size
+                logger.info(f'Downloaded {key} to {file_path} ({size} bytes)')
+            except ClientPayloadError:
+                if os.path.exists(file_path):
+                    os.remove(file_path)
+                raise
```
leplatrem: r+ :)

PR #404, opened 2018-04-13
Dump file is growing unsustainably fixes 394

peterbe:


peterbe:
@leplatrem Can you look over the latest commits after you approved. 

The reason it was failing was because the unit tests were leaking. Their ordered mattered whether the tests would pass or if one would fail.

jobs/buildhub/to_kinto.py
```diff
@@ -148,13 +234,11 @@ def done(future):
...
-        omit = ['last_modified', 'schema']
-        r1c = {k: v for k, v in r1.items() if k not in omit}
-        r2c = {k: v for k, v in r2.items() if k not in omit}
-        return r1c == r2c
-
-    records_by_id = {r['id']: r for r in existing}
+    def record_unchanged(record):
+        return (
+            record['id'] in existing and
+            existing.get(record['id']) == hash_record(record)
```
leplatrem: Since `hash_record()` never returns `None`, you can write:
```py
return existing.get(record['id']) == hash_record(record)
```
and thus even maybe drop the use of this `record_unchanged()` function

jobs/tests/test_to_kinto.py
```diff
@@ -2,37 +2,87 @@
...
+        assert not os.path.exists(self.old_cache_file)
+        with open(self.cache_file) as f:
+            records = json.load(f)
+            assert len(records) == 1
+            assert records['a'][0] == 2  # [0] is last modified
+
+        # The migration is done, but let's make sure the fetch_existing
+        # continues to work as expected.
+        mocked.get_records.return_value = [
+            {'id': 'a', 'title': 'b', 'last_modified': 2},
```
leplatrem: Since the migration is done, I suppose it should use the latest Etag, and this record does not seem unchanged, so it should not be listed here.

In order to test that the entries present in the old cache effectively end up in the new cache, it would make sense not to list this record here, so that when you make sure it is present at the end, you know it's taken from the old cache file

jobs/buildhub/to_kinto.py
peterbe: No, if I do that I have to execute `hash_record(record)` just so that it can ultimately become: `if None == 'somehashstringhere':`.

The first part of the if helps figure out if it's worth bothering calculating a hash. By doing both things the if statement becomes (if you break it down in steps): `if False and existing.get(record['id']) == hash_record(record):` which means it exists out on the first `False` and doesn't bother executing the `existing.get(record['id']) == hash_record(record)` part at all. 

Right?

leplatrem: yes

Issue #402, opened 2018-04-13
Optionally configure markus to log


peterbe:
By default markus is configured to use `markus.backends.datadog.DatadogMetrics` and it picks up `STATSD_HOST`, `STATSD_PORT` and `STATSD_NAMESPACE` as the parameters. Good. 

But when you're running things like `latest-inventory-to-kinto` locally, you get errors about not being able to send to that. You can ignore it but it's ugly and it's be much better to have the metrics redirected to logging instead.

PR #405, opened 2018-04-13
Optionally configure markus to log fixes 402

peterbe:
[See issue](https://github.com/mozilla-services/buildhub/issues/402)


docs/dev.rst
```diff
@@ -313,3 +313,17 @@ Finally, exit the interactive Docker shell and try to rebuild:
...
     $ docker-compose build buildhub
+
+Debugging metrics logging
+=========================
+
+By default, all metrics logging goes to
+``markus.backends.datadog.DatadogMetrics``. This requires that you have a
+``statsd`` server running on ``$STATSD_HOST:STATSD_PORT``. If you don't
+have that locally you can change it to plain Python logging by setting
+``LOG_METRICS=true``. For example:
```
leplatrem: Maybe it would be more explicit to have a string value, like `LOG_METRICS=python` or `LOG_METRICS=datadog` (default)

peterbe: Actually, I think that's great idea. I'll switch to that.

peterbe: What should happen if someone accidentally sets `LOG_METRICS=logggging`? I.e. a typo.

leplatrem: Yes, it's fine, even a `ValueError`

PR #411, opened 2018-04-20
Ability to quiet markus, fixes 410

peterbe:


jobs/buildhub/configure_markus.py
```diff
@@ -39,10 +40,39 @@ def get_metrics(namespace):
                     }
                 }
             ])
+        elif log_metrics_config == 'void':
+            markus.configure([
+                {
+                    'class': 'buildhub.configure_markus.VoidMetrics',
+                }
+            ])
```
willkg: Pretty sure you can just have it be an empty list--you don't need to invent a no-op metrics backend.

This should work:

```
markus.configure([])
```

Pretty sure it starts off with an empty list already, so you might even be able to just `pass` here.

PR #414, opened 2018-04-20
run fetch_existing only once, fixes #412

### LINKED ISSUE #412: fetch_existing is done twice per inventory
When you run `latest-inventory-to-kinto` it runs two parallel `main` functions. Same `loop` but different `inventory`. 
See https://github.com/mozilla-services/buildhub/blob/3a664b9c83243a3fb795277d7fa92fea04617574/jobs/buildhub/s3_inventory_to_kinto.py#L292

What then happens is that the rather memory expensive [fetch_existing()](https://github.com/mozilla-services/buildhub/blob/3a664b9c83243a3fb795277d7fa92fea04617574/jobs/buildhub/to_kinto.py#L106) is run twice. In parallel.

peterbe:


peterbe:
I've tested this locally. It works. But I'm still a bit unsure about this change since I didn't really know why it had to be done as "two separate coroutines".

peterbe:
I made it a constant instead. See https://github.com/mozilla-services/buildhub/pull/414/commits/ba00ca535c2fe9f98aa17e728b7df4f958fc768c

The only thing missing is a doc string to explain this variable.

PR #417, opened 2018-04-20
upgrade aiohttp to 3.1.3

peterbe:


peterbe:
Sigh. 59 test failures. No small change. Tempted to just park this eons.

PR #430, opened 2018-04-26
Tests shouldn't depend on the network, fixes #429

### LINKED ISSUE #429: Tests shouldn't depend on the network
The tests shouldn't have to depend on the network. It makes it unpredictable in CI. 
Actual network fetches should be caught and mocked.

peterbe:


peterbe:
@leplatrem So I changed direction here. I guess we're going to have to accept that the functional tests really do depend on archive.mozilla.org for reals. 

At least I was able to check that the unit tests don't leak. And I wrote some documentation about how to keep testing that. r?

peterbe:
Just going to land this. It's so harmless.

PR #442, opened 2018-05-01
in lambda, fetch_listing with retry_on_notfound, fixes #424

### LINKED ISSUE #424: fetch_listing in lambda_s3_event should retry
See https://sentry.prod.mozaws.net/operations/buildhub-stage/issues/4280596/

```
ClientResponseError: 404, message='Not Found'
  File "buildhub/inventory_to_records.py", line 73, in fetch_listing
    data = await fetch_json(session, url)
  File "backoff/_async.py", line 120, in retry
    ret = yield from target(*args, **kwargs)
  File "buildhub/inventory_to_records.py", line 60, in fetch_json
    response.raise_for_status()
  File "aiohttp/client_reqrep.py", line 718, in raise_for_status
    headers=self.headers)

ValueError: Could not fetch 'https://archive.mozilla.org/pub/firefox/candidates/60.0b15-candidates/build1/mac/en-US/': 404, message='Not Found'
  File "buildhub/lambda_s3_event.py", line 214, in lambda_handler
    loop.run_until_complete(main(loop, event))
  File "asyncio/base_events.py", line 466, in run_until_complete
    return future.result()
  File "buildhub/lambda_s3_event.py", line 120, in main
    _, files = await fetch_listing(session, l10n_parent_url + locale)
  File "buildhub/inventory_to_records.py", line 76, in fetch_listing
    raise ValueError("Could not fetch '{}': {}".format(url, e))
```

The [URL in the exception message](https://archive.mozilla.org/pub/firefox/candidates/60.0b15-candidates/build1/mac/en-US/) clearly doesn't 404 any more. So this should have retried.

peterbe:
I don't know if this is the only one where this can happen in the "lambda context". It's a start and should resolve https://sentry.prod.mozaws.net/operations/buildhub-stage/issues/4297460/


jobs/buildhub/inventory_to_records.py
```diff
@@ -107,9 +107,13 @@ class JSONFileNotFound(Exception):
...
     try:
-        data = await fetch_json(session, url)
+        data = await fetch_json(
+            session,
+            url,
+            retry_on_notfound=retry_on_notfound,
+        )
         return data['prefixes'], data['files']
     except (aiohttp.ClientError, KeyError, ValueError) as e:
         raise ValueError("Could not fetch '{}': {}".format(url, e))
```
leplatrem: We may want to add a test here https://github.com/peterbe/buildhub/blob/6ac3fa0f1708558f55faeb5a205fd66bf975dbf4/jobs/tests/test_inventory_to_records.py#L192 ;)

peterbe: I'm actually not a fan of this. It's not great that it lumps together all of these different exceptions. 

HTTP problems should probably be separated from problems with the content. The HTTP problems might sometimes be something you want to retry but if the content is busted it will never help to retry. 

Let we discussed. Hopefully this PR will at least solve one of the tracebacks we saw in Sentry.

PR #444, opened 2018-05-02
Use ciso8601 to parse CSV dates, fixes #443

### LINKED ISSUE #443: Use ciso8601 to parse CSV dates
We do about 50M of these every day and [ciso8061 is apparently 13 times faster](https://www.peterbe.com/plog/fastest-python-datetime-parser).

peterbe:


peterbe:
The reason is because I had `pytz` installed in my virtualenv. It's not in requirements.txt so I must have installed it from some other dev tool. 

See https://github.com/closeio/ciso8601/issues/19

Issue #446, opened 2018-05-02
Generating existing from scratch is expensive


peterbe:
Suppose you delete the `.records-hashes-$SERVER_NAME.json` file. That means this code right here: https://github.com/mozilla-services/buildhub/blob/ee484fab5ac852036b355da0d8be1910955ee3b9/jobs/buildhub/to_kinto.py#L158-L161 is going to extract every single record from the Kinto database into a monster list of dicts. 

I measured it, with `memory_profiler` in my docker with a postgres database of about 500,000 records. Generating that `new_records` dict eats up about 1700MB.
See https://irccloud.mozilla.com/pastebin/ZhAoBDia/

The solution has to be to consume that paginated result as a stream and for each patch of 10,000 records make the hashes and then reuse that allocated memory for the next 10,000 records. 

@leplatrem I don't see a whole lot in [the documentation](https://github.com/Kinto/kinto-http.py#pagination) about to do the pagination i any other way. Do you have some ideas?

bqbn:
Did some manual runs on -stage using v1.2.1, and it shows that the program requires at least 5 GB memory in order to finish, or it gets OOM killed.

Without this being fixed, we can expect that the program will need more memory due to the ever increasing database.

leplatrem:
> @leplatrem I don't see a whole lot in the documentation about to do the pagination i any other way. Do you have some ideas?

Indeed, the pagination is not exposed in the kinto-http client :(

As a work around, using raw requests is relatively easy (*untested*):

```py

url = records_url
while True:
   resp = requests.get(url, params={"_since": timestamp, "_limit": 1000})
   page_records = resp.json["data"]
   # ... process(page_records)
   url = resp.headers.get("Next-Page")
   if url is None:
       break

```

peterbe:
Fixed by https://github.com/mozilla-services/buildhub/commit/71f34fd67635c9bde3e6b6e5668a34838c4f6fe2

Issue #445, opened 2018-05-02
Remove to_kinto.fetch_existing dump file migration code


peterbe:
In https://github.com/mozilla-services/buildhub/commit/dd361a334fcd5a6d303598e3ff3cda74b1397caa we totally changed the way the dump file works. Instead of a big list of dicts we just store a dict of `{ID: stringhash}`.

Included in that PR was a migration script that would "copy" the old `.records-$SERVERNAME.json` file to the new `.records-hashes-$SERVERNAME.json` file. 

That was dumb because when you make a new release to upgrade Stage or Prod, the EC2 is completely recreated. The new server replaces the old server so there's no disk cache to migrate from.

peterbe:
Fixed by https://github.com/mozilla-services/buildhub/commit/71f34fd67635c9bde3e6b6e5668a34838c4f6fe2

PR #448, opened 2018-05-03
optimize fetch_existing, fixes #447 #445 #446

### LINKED ISSUE #447: No need to fear mutating existing records for dump file
The purpose of calling `fetch_existing` is to generate a massive Python dict of `{ID: stringhash}`. We're currently doing this in a way that does NOT mutate the record (when it comes from Kinto). 
See https://github.com/mozilla-services/buildhub/blob/ee484fab5ac852036b355da0d8be1910955ee3b9/jobs/buildhub/to_kinto.py#L166
The not-mutating is because we call [`hash_record`](https://github.com/mozilla-services/buildhub/blob/ee484fab5ac852036b355da0d8be1910955ee3b9/jobs/buildhub/to_kinto.py#L83-L89) which does a deep copy (expensive!)

peterbe:
WIP. I haven't added a new test for the iterator version of getting new_records.


peterbe:
CHECK THIS OUT!! @bqbn @leplatrem 
https://gist.github.com/peterbe/1ab8ba3b97db71a159db9e5452b06e67

First of all, I don't understand the "Increment" column. What the heck is `-8132224.9 MiB`? Negative 8 Terabytes?

Anyway, the total memory usage for the whole `fetch_existing` function is now 340MiB! Last night when I ran this against my local Kinto using the old technique it increased by 1700MiB instead. 

Also, this PR has some other cool features:

1. A statsd counter of how many records are extracted inside `fetch_existing`. 
2. A statsd timer on executing the whole function
3. Got rid of the old db dump migration code. 
4. Using the faster, but mutating, function of turning each existing kinto record into a hash string.

jobs/buildhub/to_kinto.py
```diff
@@ -155,16 +102,53 @@ def fetch_existing(
...
-        _since=previous_run_etag,
-        pages=float('inf')
-    )
-
-    for record in new_records:
-        records[record['id']] = [
-            record['last_modified'],
-            hash_record(record)
-        ]
+    # The reason we can't use client.get_records() is because it is not
```
leplatrem: If you want you can add a mention of https://github.com/Kinto/kinto-http.py/issues/175

jobs/buildhub/to_kinto.py
```diff
@@ -155,16 +102,53 @@ def fetch_existing(
...
+        # However, we can use it if there was a previous_run_etag which
+        # means we only need to extract a limited amount of records. Not
+        # the whole Kinto database.
+        new_records_batches = [client.get_records(
+            _since=previous_run_etag,
+            pages=float('inf')
+        )]
+    else:
+        def new_records_iterator():
+            params = {'_since': None}
```
leplatrem: params is thus useless, no?

jobs/buildhub/to_kinto.py
```diff
@@ -155,16 +102,53 @@ def fetch_existing(
...
+            next_url = None
+            while True:
+                if next_url is None:
+                    endpoint = client.get_endpoint('records')
```
leplatrem: why have endpoint and next_url variables instead of just one initialized out of the loop?

jobs/buildhub/to_kinto.py
```diff
@@ -155,16 +102,53 @@ def fetch_existing(
...
+        new_records_batches = new_records_iterator()
+
+    count_new_records = 0
+    for new_records in new_records_batches:
+        for record in new_records:
+            count_new_records += 1
+            records[record['id']] = [
+                record['last_modified'],
+                hash_record_mutate(record)
+            ]
```
leplatrem: :+1:

jobs/buildhub/to_kinto.py
peterbe: I don't understand your comment. It's set to `{'_since': None}` for the first request. Then, for all the other paginated requests it's `{}`.

jobs/buildhub/to_kinto.py
peterbe: I refactored it a little so there's just one variable name that gets reused. 

The `endpoint` (when I print it) looks like this:
```
ENDPOINT /buckets/build-hub/collections/releases/records
ENDPOINT http://localhost:8888/v1/buckets/build-hub/collections/releases/records?_limit=10000&_token=eyJsYXN0X3JlY29yZCI6eyJsYXN0X21vZGlmaWVkIjoxNTI1MzY1MzUyODkyfSwib2Zmc2V0IjoxMDAwMCwibm9uY2UiOiJwYWdpbmF0aW9uLXRva2VuLWZlMjJlNzdmLTg0OWQtNGRjOC1hOTI5LTI4NDNjMWY2MWUzOSJ9
ENDPOINT http://localhost:8888/v1/buckets/build-hub/collections/releases/records?_limit=10000&_token=eyJsYXN0X3JlY29yZCI6eyJsYXN0X21vZGlmaWVkIjoxNTI0NTg5MjY2MjU3fSwib2Zmc2V0IjoyMDAwMCwibm9uY2UiOiJwYWdpbmF0aW9uLXRva2VuLWIyZjA4OTU5LTU4NTgtNGMzNC1iOGUxLWI0Yjc0Zjk2Njg3MSJ9
ENDPOINT http://localhost:8888/v1/buckets/build-hub/collections/releases/records?_limit=10000&_token=eyJsYXN0X3JlY29yZCI6eyJsYXN0X21vZGlmaWVkIjoxNTI0MjQxMzYwNTI5fSwib2Zmc2V0IjozMDAwMCwibm9uY2UiOiJwYWdpbmF0aW9uLXRva2VuLTZkNjQ3NGMwLTkzMTctNDYwNS05ZjU2LTQ1OTg2NWFhYmUxYSJ9
...
```

PR #452, opened 2018-05-07
insert buildhub release into raven config

peterbe:
The purpose of this is to indicate to Sentry to distinguish between exceptions for different releases. 
The version of `buildhub` is not the same as the release tag in git, but it happens to be the same numbers. 

I tested this locally by running: `./peterbe_execute_lambda.py jobs/tests/data/s3-event-simple.json` and it correctly set the value to `'1.3.0'`.


peterbe:
> I would have preferred if it was defined once in buildhub.__version__

Good catch!

Updated now.

PR #467, opened 2018-05-15
it's lambda not lamdba

peterbe:
More important PR of the year @leplatrem


peterbe:
FYI, the reason Circle is failing is because of Dockerhub network status.

PR #469, opened 2018-05-17
Lambda misses builds with spaces, fixes #468

### LINKED ISSUE #468: Lambda function misses certain builds because of spaces
@wlach discovered that when Firefox 60.0.1 was built yesterday only the `['linux-i686', 'linux-x86_64']` platforms were saved in Buildhub. Curious. 

Compare: [Windows](http://archive.mozilla.org/pub/firefox/releases/60.0.1/win32/en-US/) and [Linux](http://archive.mozilla.org/pub/firefox/releases/60.0.1/linux-i686/en-US/). 

One thing to notice is that the Windows filename contains spaces, the Linux one does not. The reason we think the cron job does not have a problem with this is because of that space. See [this code from the cron job](https://github.com/mozilla-services/buildhub/blob/fd019fbd31699d6a7be3a567c8c66ec66be73175/jobs/buildhub/inventory_to_records.py#L476).

peterbe:


peterbe:
@leplatrem I don't know if you have time to look into this but I'm basing this on a "hunch". 
It just seemed odd that when going through the inventory the keys might have plusses in them and unless we replace those with spaces, it won't work. 

What happened to us was that only the linux platform ones worked via Lambda. No errors in Sentry. The Mac and Windows only worked when we eventually used the cron job.

PR #527, opened 2018-11-15
fixes bug 1506862 - linux/mac nightlies missing since nov 5th

peterbe:


jobs/buildhub/inventory_to_records.py
```diff
@@ -443,9 +420,21 @@ def deduplicate_entries(entries):
...
+                # known FILE_EXTENSIONS it will never pass a build URL
+                # later in the loop.
+                if not any(
+                    object_key.endswith(ext) for ext in FILE_EXTENSIONS
+                ):
+                    # Actually, eventually that FILE_EXTENSION check will
+                    # be done again more detailed inside the is_build_url
+                    # function.
+                    # This step was just to weed out some easy ones.
+                    continue
```
peterbe: THESE lines are the most important change. Basically instead of trying to be "too smart" it just does the simplest and fastest of checks. If the file has a decent file extension it gets to participate. Otherwise it's skipped. 

A good thing about this change too is that it removes the need to A) create new list in memory B) create another list from calling `sorted()` and C) not having to create a dict. 

Here's how you can "test" it...

See this file:
https://gist.github.com/peterbe/0b0fb510f3b2b49ab4657db03a83cc9b
If you download it and run it with this horrible script:

```python
import json
from buildhub.utils import key_to_archive_url, is_build_url

PRODUCTS = ('firefox',)
with open('2018-11-13-10-00-51-mozilla-central.entries.log') as f:
    entries = json.load(f)


for thing in entries:
    object_key = thing['Key']
    try:
        # /pub/thunderbird/nightly/...
        product = object_key.split('/')[1]
    except IndexError:
        continue  # e.g. https://archive.mozilla.org/favicon.ico
    if product not in PRODUCTS:
        continue
    url = key_to_archive_url(object_key)
    print(bool(is_build_url(product, url)), url)
```
The output I get is this: https://gist.githubusercontent.com/peterbe/05d803f6830c2292b6dd3d34a2556f6b/raw/d00b2bd5c8549f6f15c1ccddb82c0075e5f003ea/output.log

jobs/buildhub/inventory_to_records.py
```diff
@@ -485,6 +468,10 @@ def deduplicate_entries(entries):
                     logger.exception(e)
                     continue
 
+                # Scan the list of candidates metadata (no-op if
+                # already initialized).
+                await scan_candidates(session, product)
```
peterbe: I just moved these lines to be AFTER the repeatedly called `is_build_url()`. It's a "no-op if already initialized" but it's still an `await` which means it gives the loop a chance to run other things.

jobs/buildhub/s3_inventory_to_kinto.py
```diff
@@ -64,6 +64,10 @@
     cast=Csv()
 ))
 
+LOG_LEVEL = config('LOG_LEVEL', default='INFO')
+
+STORE_DAILY_MANIFEST = config('STORE_DAILY_MANIFEST', default=False, cast=bool)
```
peterbe: I just needed this for local dev. It can be ignored.

peterbe: What I found was weird. 

I would run the `latest-inventory-to-kinto` once (takes 1-2 hours) and it downloads all the `md5hashmd5hashmd5hashmd5hashmd5hash.csv.gz` files to disk. But when I run it immediately after, the code would download the inventory manifest again and now it's finding it needs to download new .csv.gz files. All within a single day!

jobs/buildhub/utils.py
```diff
@@ -150,6 +155,22 @@ def build_record_id(record):
...
+    '.+(tinderbox|try-builds|partner-repacks|latest|contrib|/0\.|'
+    'experimental|namoroka|debug|sha1-installers|candidates/archived|'
+    'stylo-bindings|/1.0rc/|/releases/win../|dominspector|/test/|testing|'
+    '%28.+%29|\sInstaller\.(\w{2,3}\-?\w{0,3})\.exe)'
+)
+
+_build_url_exclude_suffixes_regex = re.compile(
+    '.+(sdk|tests|crashreporter|stub|gtk2.+xft|source|asan)'
+)
+
```
peterbe: I just moved these outside the function since they're the same every time. This way the `re.compile` can be called *once* for the whole scraper run instead of thousands of times.

jobs/buildhub/utils.py
```diff
@@ -177,36 +198,27 @@ def is_build_url(product, url):
...
-    if re_exclude.match(url):
+    if _build_url_exclude_names_regex.match(url):
         return False
 
+    extensions = FILE_EXTENSIONS
     # Only .exe for Windows.
-    extensions = list(FILE_EXTENSIONS)
     if 'win' in url:
-        extensions.remove('zip')
+        extensions = [x for x in extensions if x != 'zip']
```
peterbe: Before, it would create a new list object (by cloning the tuple) every single time. Even though it sometimes didn't need to remove `'zip'`.

jobs/buildhub/utils.py
```diff
@@ -177,36 +198,27 @@ def is_build_url(product, url):
...
 
     if product == 'devedition':
         product = 'firefox'
     if product == 'mobile':
         product = 'fennec'
     filename = os.path.basename(url)
     match_filename = filename.replace(' ', '-').lower()
+    if _build_url_exclude_suffixes_regex.match(match_filename):
+        return False
```
peterbe: These lines used to be on the last line before. I just moved them up in case it can cause an early exit and thus we don't need to make the last regex stuff after this.

peterbe:
As a general comment, a big difference here is that the `is_build_url` function is going to be called A LOT more. Before we did a bunch of deduplications of the `entries` which spared `is_build_url`. 

Turns out, I measured `is_build_url` before my optimizations (some early exists, pre-compiled regexes) and after. It made virtually no difference. A tiny bit but immaterial compared to other things. Each call takes about 0.027 seconds.

jobs/buildhub/inventory_to_records.py
leplatrem: I don't understand do you deduplicate then?

If I remember we had cases where some folder would contain:
- *.exe, *.zip, *.installer.exe 
- only *.installer.exe
- only *.zip

leplatrem: You may have been to reproduce the bug you describe by adding some folder entries in this file: https://github.com/mozilla-services/buildhub/blob/master/jobs/tests/data/inventory-simple.csv
For example:
```
pub/firefox/nightly/2018/11/2018-11-13-10-00-51-mozilla-central/firefox-65.0a1.en-US.win32.zip
pub/firefox/nightly/2018/11/2018-11-13-10-00-51-mozilla-central/firefox-65.0a1.en-US.win32.exe
pub/firefox/nightly/2018/11/2018-11-13-10-00-51-mozilla-central/firefox-65.0a1.en-US.win32.installer.exe
pub/firefox/nightly/2018/11/2018-11-13-10-00-51-mozilla-central/firefox-65.0a1.en-US.mac.dmg
pub/firefox/nightly/2018/11/2018-11-13-10-00-51-mozilla-central/firefox-65.0a1.en-US.linux-x86_64.tar.bz2
```
and make it would return:
- `firefox-65.0a1.en-US.win32.exe`
- `firefox-65.0a1.en-US.mac.dmg`
- `firefox-65.0a1.en-US.linux-x86_64.tar.bz2`

jobs/buildhub/inventory_to_records.py
```diff
@@ -403,6 +403,23 @@ class JSONFileNotFound(Exception):
...
+        ones that contain "setup" or "installer".
+        """
+        deduplicated = {}
+        for entry in entries:
+            identifier = entry['Key'].lower() \
+                                     .replace('+setup+', '-') \
+                                     .replace('.installer.exe', '') \
+                                     .replace('.exe', '')
+            dupped = deduplicated.get(identifier, {"Key": ""})
+            is_installer = 'installer' in dupped['Key'] or '+setup+' in dupped['Key']
```
peterbe: This will be true for `....installer-stub.exe` and `installer.tar.gz` too. 
Not sure that that means. Just makes me nervous.

leplatrem: installer-stub will be filtered by is_build_url() I believe

leplatrem: You are right, I will adjust this

Issue #532, opened 2019-06-10
THIS PROJECT IS DEPRECATED


willkg:
Buildhub (this project) is deprecated and is no longer actively supported.

Buildhub will be decommissioned soon--possibly July 2019. (Actual schedule TBD.)

If you are using Buildhub, please migrate to using Buildhub2: https://buildhub.moz.tools/

willkg:
* [] add issue to issue tracker
* [] add deprecation notice to web ui
* [] edit GitHub project summary and README
* [] email stability mailing list
* [] write blog post that's syndicated on Planet Mozilla
* [] add issues to issue trackers for projects we know are using Buildhub

willkg:
https://bugzilla.mozilla.org/show_bug.cgi?id=1559004 covers fixing Socorro buildid links.

willkg:
I talked with @marco-c and @calixteman about the projects they manage. I think some of them aren't used anymore. I'm getting that clarified.

I'm probably going to fix some of them. I'm working on those now.

My goal is to have everything figured out by the end of next week so we can decom Buildhub sooner rather than later.

willkg:
Status July 4th, 2019:

* I think yorstats isn't in use anymore.
* I did a PR for crash-clouseau which hasn't landed: https://github.com/mozilla/crash-clouseau/pull/7
* I did a PR for pushlog-addon which hasn't landed: https://github.com/mozilla/pushlog-addon/pull/8
* I'm pretty sure crashstop isn't in use anymore. crash-stop-addon is fixed.
* firefox-workflow-analysis isn't in use anymore.

Given that, I think we're good to go for decommissioning Buildhub.

marco-c:
@calixteman can you archive the repos which are not used anymore?

marco-c:
BTW, thanks for all the PRs, @willkg!

PR #534, opened 2019-06-10
Add deprecation notice

willkg:
This is for issue #532


willkg:
There's probably a better way to do this, but sticking it in the result section was easy and a minimal change, so I did it this way. Probably didn't need to turn it into a component, but figured I'd do that in case I needed it elsewhere, too.

self-r+

PR #535, opened 2019-06-10
Fix release process

willkg:
This fixes bin/make-release.py to run in a docker container.


willkg:
I think this will work. We'll see. self-r+

